PR #5, opened 2023-07-26
example script demonstrating how to use pget asynchronously 

technillogue:
the idea is that this starts downloading before importing torch etc, while also being robust against predict.py being imported multiple times

### DIFF
--- a/pget_util.py
+++ b/pget_util.py
@@ -1,0 +1,73 @@
+import os
+import pathlib
+import subprocess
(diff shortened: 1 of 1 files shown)


technillogue:
in retrospect the correct way to do this is 

```
import multiprocessing as mp

if mp.current_process().name != "MainProcess":
     download()
```

but that's not as relevant anymore

PR #9, opened 2023-08-04
WIP: use less memory by downloading to sparse file

philandstuff:
pget has a problem that it needs to download everything into memory along each of the different parallel connections before finally saving everything to disk.  This has two consequences:

- pget needs to consume at least as much memory as the size of the downloaded file
- pget cannot begin writing to disk until all the bytes have been downloaded from source

I had a thought that we could precreate a sparse file of the correct length by using the os.Truncate(size) call.  Then each goroutine can write to different offsets within the same file, in parallel.

I tested this locally on macOS and I was able to download a 308MB file with pget's memory usage only reaching 26MB.

This commit is a WIP because I've commented out the untar stuff to keep the proof of concept simple.  We should restore (or deprecate?) that functionality before merging this.  We should also do more extensive performance testing.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,7 +6,6 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


technillogue:
very cute use of truncate

philandstuff:
Another thing we might want to do here before it gets merged is write to a `.partial` file and only rename to the final dest filename once we've downloaded all the bytes, to avoid leaving a half-downloaded sparse file in the wrong place if pget gets interrupted.

philandstuff:
I've reimplemented extract-tar and done the `.partial` thing. Ready for review.

main.go
```diff
@@ -195,26 +214,44 @@ func main() {
...
 	}
 
-	buffer, err := downloadFileToBuffer(url, *concurrency)
+	// create tempfile for downloading to
+	cwd, err := os.Getwd()
+	if err != nil {
+		fmt.Printf("Error getting cwd: %v\n", err)
+		os.Exit(1)
+	}
+	destTemp, err := os.CreateTemp(cwd, dest+".partial")
```
tempusfrangit, line 223: Does it make sense to use the `*` notation to allow createTemp to create a unique file for the temp, this means that if somehow something is left on disk afterwards/failure of pget before removal, a new temp file would be created instead of trying to scribble over the top

```suggestion
	destTemp, err := os.CreateTemp(cwd, fmt.Sprintf("%s-*.partial", dest)
```

or even `os.CreateTemp(cwd, "*.partial")` which it more-inline with how browsers do it now, so that filenames are really obscured and not accidentally used until complete. Since we either extract from the reference object for the tempfile or rename it, the name doesn't need to include "dest" if we don't want it.

tempusfrangit (approved):
One question, nothing blocking for sure.

I also wonder if it's worth maintaining a "toBuffer" form at all? For the record, I really like the "scribble to disk to avoid OOMing pget"

tempusfrangit:
Also, related but not blocking, we should implement testing :)

PR #17, opened 2023-11-01
handle small files better

technillogue:
- limit concurrency for small files
- potentially let others know how many pgets are running

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,7 +6,6 @@ import (
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


anotherjesse:
@technillogue this seems reasonable. 

I was originally thinking we could set a "minChunkSize" - with a default of ~10-100MB? (override via args?)

If we want to do that, we won't need this logic.

this is good - 👍  - or if you want to do the chunk approach we can just do min chunk

PR #21, opened 2023-11-13
Restructure download

tempusfrangit:
* Custom Transport (roundtripper)
* Use an errgroup not a waitgroup
* Version Command
* user-agent string

### DIFF
--- a/download/buffer.go
+++ b/download/buffer.go
@@ -2,15 +2,16 @@ package download
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"math"
 	"net/http"
-	"sync"
 	"time"
 
 	"github.com/dustin/go-humanize"
 	"github.com/spf13/viper"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/replicate/pget/optname"
 )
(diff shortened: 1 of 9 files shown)


tempusfrangit:
STACKED COMMIT -> Stacked with  #20 (merge #20 first)

anotherjesse:
I've not yet tested this, but overall this LGTM

download/buffer.go
```diff
@@ -85,88 +73,43 @@ func DownloadFileToBuffer(url string) (*bytes.Buffer, int64, error) {
...
 	elapsed := time.Since(startTime).Seconds()
 	througput := humanize.Bytes(uint64(float64(fileSize) / elapsed))
 	fmt.Printf("Downloaded %s bytes in %.3fs (%s/s)\n", humanize.Bytes(uint64(fileSize)), elapsed, througput)
 
 	buffer := bytes.NewBuffer(data)
 	return buffer, fileSize, nil
 }
+
+func downloadChunk(ctx context.Context, start, end int64, dataSlice []byte, trueURL string) error {
+	client := newClient()
```
technillogue, line 94: recreating the client for each chunk always bothered me but we can handle this once we get to downloading multiple files

download/buffer.go
```diff
@@ -0,0 +1,115 @@
...
+		concurrency = maxConcurrency
+		chunkSize = fileSize / int64(concurrency)
+	}
+	if viper.GetBool(optname.Verbose) {
+		fmt.Printf("Downloading %s bytes with %d connections (chunk-size = %s)\n", humanize.Bytes(uint64(fileSize)), concurrency, humanize.Bytes(uint64(chunkSize)))
+	}
+
+	errGroup, ctx := errgroup.WithContext(context.Background())
+
+	data := make([]byte, fileSize)
```
technillogue, line 65: we'll need to put a lot of thought into this line

technillogue (approved):

download/buffer.go
tempusfrangit: Part of my plan is to evolve so we have a few new ways to download and we can keep the default as is. This buffer creation bit is just the current way we do things. I def. want something better than "ALLOCATE ALL THE RAM" :)

tempusfrangit:
Tested locally, this appears to work as expected.

PR #24, opened 2023-11-14
Utility argument additions

tempusfrangit:
* --resolve - allows for curl-like definitions of targets to resolve for a given hostname
* --connect-timeout - allows to set the connection timeout

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,10 +6,12 @@ import (
 	"github.com/spf13/viper"
 	"runtime"
 	"strings"
+	"time"
 )
 
 var (
 	Concurrency      int
+	ConnTimeout      time.Duration
 	Extract          bool
 	Force            bool
 	MinimumChunkSize string
@@ -19,6 +21,7 @@ var (
 
 func AddFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().IntVarP(&Concurrency, optname.Concurrency, "c", runtime.GOMAXPROCS(0)*4, "Maximum number of concurrent downloads")
+	cmd.PersistentFlags().DurationVar(&ConnTimeout, optname.ConnTimeout, 5*time.Second, "Timeout for establishing a connection, format is <number><unit>, e.g. 10s")
 	cmd.PersistentFlags().BoolVarP(&Extract, optname.Extract, "x", false, "Extract archive after download")
 	cmd.PersistentFlags().StringVarP(&MinimumChunkSize, optname.MinimumChunkSize, "m", "16M", "Minimum chunk size (in bytes) to use when downloading a file (e.g. 10M)")
 	cmd.PersistentFlags().BoolVarP(&Force, optname.Force, "f", false, "Force download, overwriting existing file")
--- a/config/config.go
+++ b/config/config.go
@@ -58,16 +58,18 @@ func convertResolveHostsToMap() error {
 		if len(split) != 3 {
 			return fmt.Errorf("invalid resolve host format, expected <hostname>:port:<ip>, got: %s", resolveHost)
 		}
-		host, port, addr := strings.ToLower(split[0]), split[1], split[2]
-		hostPort := fmt.Sprintf("%s:%s", host, port)
-		if _, ok := HostToIPResolutionMap[hostPort]; ok {
-			return fmt.Errorf("duplicate host specified: %s", host)
+		host, port, addr := split[0], split[1], split[2]
+		if net.ParseIP(host) != nil {
+			return fmt.Errorf("invalid hostname specified, looks like an IP address: %s", host)
 		}
-		IP := net.ParseIP(addr)
-		if IP == nil {
+		hostPort := net.JoinHostPort(host, port)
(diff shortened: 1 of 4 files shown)


technillogue:
what's the usecase for this? I'd like to keep the feature surface to only things that we actually need to use

tempusfrangit:
> what's the usecase for this? I'd like to keep the feature surface to only things that we actually need to use -- I'd rather not have stuff only used for testing, and avoiding DNS in prod doesn't seem like enough of an improvement

This is extremely useful for things like storagecache warming and any time we need to validate a download against <target>; example I have weights stood up locally and do not want to download it, but I want to go through the same path as I normally would I can specify weights.replicate.delivery:443:127.0.0.1 (even in ENV) and validate the entire path works as expected.

This is general utility things and will be helpful for expanded use-cases.

config/config.go
```diff
@@ -32,3 +42,39 @@ func AddFlags(cmd *cobra.Command) {
...
+		if _, ok := HostToIPResolutionMap[host]; ok {
+			return fmt.Errorf("duplicate hostname specified: %s", host)
+		}
+		IP := net.ParseIP(addr)
+		if IP == nil {
+			return fmt.Errorf("invalid IP address: %s", addr)
+		}
+		if IP.To16() != nil && IP.To4() == nil {
+			return fmt.Errorf("invalid IP address not IPv4 or IPv6: %s", addr)
+		}
```
philandstuff, line 71:
```suggestion
		if IP.To16() == nil && IP.To4() == nil {
			return fmt.Errorf("invalid IP address not IPv4 or IPv6: %s", addr)
		}
```
I think?

Although I don't know why we need this check - if net.ParseIP was successful, we don't need to check whether it's v4 or v6?

download/client.go
```diff
@@ -92,13 +95,25 @@ func newClient() *http.Client {
...
 	return nil
 }
 
+// transportDialContext is a wrapper around net.Dialer that allows for overriding DNS lookups via the values passed to
+// `--resolve` argument.
 func transportDialContext(dialer *net.Dialer) func(context.Context, string, string) (net.Conn, error) {
-	return dialer.DialContext
+	// Allow for overriding DNS lookups in the dialer without impacting Host and SSL resolution
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+		if addrOverride := config.HostToIPResolutionMap[dialer.LocalAddr.String()]; addrOverride != "" {
```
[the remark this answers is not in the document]
philandstuff: Yes, from [reading the docs](https://pkg.go.dev/net#Dial) `addr` is exactly what we want here.
```suggestion
		if addrOverride := config.HostToIPResolutionMap[addr]; addrOverride != "" {
```

config/config.go
```diff
@@ -32,3 +42,39 @@ func AddFlags(cmd *cobra.Command) {
...
+
+}
+
+func convertResolveHostsToMap() error {
+	for _, resolveHost := range viper.GetStringSlice(optname.Resolve) {
+		split := strings.SplitN(resolveHost, ":", 3)
+		if len(split) != 3 {
+			return fmt.Errorf("invalid resolve host format, expected <hostname>:port:<ip>, got: %s", resolveHost)
+		}
+		host, port, addr := strings.ToLower(split[0]), split[1], split[2]
```
philandstuff, line 61: I don't like the ToLower here, I don't think we need it

config/config.go
```diff
@@ -32,3 +42,39 @@ func AddFlags(cmd *cobra.Command) {
...
+		if _, ok := HostToIPResolutionMap[host]; ok {
```
philandstuff, line 62: we don't store `host` in the map, so `ok` won't ever be true here

tempusfrangit: Hmm. It looks like this code shifted between two branches. Let me re-look at it. This is a mis-mash of two distinct iterations. Thats why this all doesn't line up.

config/config.go
philandstuff: I'd rather join `host` and `port` back into `addr` asap:
```suggestion
		addr, ipString := fmt.Sprint("%s:%s", strings.ToLower(split[0]), split[1]), split[2]
```
also it should be `addr` because that's what net.Dial calls it, so the other thing should be ip

config/config.go
```diff
@@ -32,3 +42,37 @@ func AddFlags(cmd *cobra.Command) {
...
+}
+
+func convertResolveHostsToMap() error {
+	for _, resolveHost := range viper.GetStringSlice(optname.Resolve) {
+		split := strings.SplitN(resolveHost, ":", 3)
+		if len(split) != 3 {
+			return fmt.Errorf("invalid resolve host format, expected <hostname>:port:<ip>, got: %s", resolveHost)
+		}
+		host, port, addr := strings.ToLower(split[0]), split[1], split[2]
+		hostPort := fmt.Sprintf("%s:%s", host, port)
```
philandstuff, line 62:
```suggestion
		hostPort := net.JoinHostPort(host, port)
```
https://pkg.go.dev/net#JoinHostPort

philandstuff (approved):

PR #25, opened 2023-11-14
Implement Download modes

tempusfrangit:
Add the download mode framework with `buffer` and `tar-extract` to start. This allows us to expand the modes we want to operate in without impacting defaults or other users.

### DIFF
--- a/main.go
+++ b/main.go
@@ -11,7 +11,6 @@ import (
 	"github.com/replicate/pget/cmd"
 	"github.com/replicate/pget/pkg/config"
 	"github.com/replicate/pget/pkg/download"
-	"github.com/replicate/pget/pkg/extract"
 	"github.com/replicate/pget/pkg/optname"
 )
 
@@ -57,23 +56,6 @@ func execFunc(cmd *cobra.Command, args []string) error {
 	_ = os.WriteFile(tmpFile, []byte(""), 0644)
 	defer os.Remove(tmpFile)
 
-	buffer, fileSize, err := download.FileToBuffer(url)
-	if err != nil {
-		return fmt.Errorf("error downloading file: %w", err)
-	}
-
-	// extract the tar file if the -x flag was provided
-	if viper.GetBool(optname.Extract) {
-		err = extract.ExtractTarFile(buffer, dest, fileSize)
-		if err != nil {
-			return fmt.Errorf("error extracting file: %v", err)
-		}
-	} else {
-		// if -x flag is not set, save the buffer to a file
-		err = os.WriteFile(dest, buffer.Bytes(), 0644)
-		if err != nil {
-			return fmt.Errorf("error writing file: %v", err)
-		}
-	}
-	return nil
+	mode := download.GetMode(config.Mode)
+	return mode.DownloadFile(url, dest)
 }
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -32,3 +46,45 @@ func AddFlags(cmd *cobra.Command) {
...
+		if _, ok := HostToIPResolutionMap[host]; ok {
+			return fmt.Errorf("duplicate hostname specified: %s", host)
+		}
+		IP := net.ParseIP(addr)
+		if IP == nil {
+			return fmt.Errorf("invalid IP address: %s", addr)
+		}
+		if IP.To16() != nil && IP.To4() == nil {
+			return fmt.Errorf("invalid IP address not IPv4 or IPv6: %s", addr)
+		}
```
philandstuff, line 81:
```suggestion
		if IP.To16() == nil && IP.To4() == nil {
			return fmt.Errorf("invalid IP address not IPv4 or IPv6: %s", addr)
		}
```
I think?

Although I don't know why we need this check - if net.ParseIP was successful, we don't need to check whether it's v4 or v6?

tempusfrangit: I think you're right.. We can probably drop the suspenders and a belt.

download/client.go
```diff
@@ -92,13 +95,25 @@ func newClient() *http.Client {
...
 	return nil
 }
 
+// transportDialContext is a wrapper around net.Dialer that allows for overriding DNS lookups via the values passed to
+// `--resolve` argument.
 func transportDialContext(dialer *net.Dialer) func(context.Context, string, string) (net.Conn, error) {
-	return dialer.DialContext
+	// Allow for overriding DNS lookups in the dialer without impacting Host and SSL resolution
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+		if addrOverride := config.HostToIPResolutionMap[dialer.LocalAddr.String()]; addrOverride != "" {
```
philandstuff, line 112: This looks wrong? I would expect HostToIPResolutionMap contains strings of the form `example.com:443` and dialer.LocalAddr.String() is of the form `0.0.0.0` or `127.0.0.1`.

I think we should be using `addr` here instead maybe?

config/config.go
```diff
@@ -45,6 +47,12 @@ func AddFlags(cmd *cobra.Command) {
 
 func PersistentStartupProcessFlags() error {
 
+	if viper.GetBool(optname.Extract) {
+		if viper.GetString(optname.Mode) != "buffer" || viper.GetString(optname.Mode) != "tar-extract" {
```
philandstuff, line 53: There is no way for this condition to be true. I think you want && not ||

download/buffer.go
```diff
@@ -16,34 +17,43 @@ import (
 	"github.com/replicate/pget/optname"
 )
 
-var (
-	fileSize int64
-)
+const bufferHelpString = `
+Buffer mode downloads the file into a memory buffer before writing it to disk. This mode is useful for downloading
+files smaller than the available memory on the system.
```
philandstuff, line 22: This help text leaves me wondering "what if I want to download bigger files?"

philandstuff (approved):

pkg/optname/names.go
```diff
@@ -3,6 +3,7 @@ package optname
 const (
 	Concurrency      = "concurrency"
 	ConnTimeout      = "connect-timeout"
+	Mode             = "mode"
```
philandstuff, line 6: is this still used?

tempusfrangit: Nope.

PR #28, opened 2023-11-15
Implement Multifile

tempusfrangit:
Implements multifile command for pget.

This will take input on STDIN or a filename. EOF will start the process.


cmd/multifile.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+const multiFileLongDesc = `
+'multifile' mode for pget takes a manifest file as input (can use '-' for stdin) and downloads all files listed in the manifest.
+
+The manifest is expected to be in the format of a newline-separated list of pairs of URLs and destination paths, separated by a space.
+e.g.
+https://example.com/file1.txt /tmp/file1.txt
+
+'multifile'' will download files in parallel limited to the '--maximum-connections-per-host' limit for per-host limts and
```
philandstuff, line 26:
```suggestion
'multifile' will download files in parallel limited to the '--maximum-connections-per-host' limit for per-host limts and 
```

cmd/multifile.go
```diff
@@ -0,0 +1,182 @@
...
+		if err != nil {
+			return fmt.Errorf("error opening manifest file %s: %w", manifestPath, err)
+		}
+		defer manifestFile.Close()
+		reader = bufio.NewReader(manifestFile)
+	}
+	// process the manifest file into a map of hosts to url/file pairs with processManifest
+	buffer := make([]string, 0)
+	for {
+		readString, err := reader.ReadString('\n')
```
philandstuff, line 96: we could also use a bufio.Scanner here https://pkg.go.dev/bufio#Scanner

philandstuff: though I think this would be simpler if we did os.ReadFile and just operated on the whole manifest file in memory https://pkg.go.dev/os#ReadFile

cmd/multifile.go
```diff
@@ -0,0 +1,182 @@
...
+		// add the url to seenURLs
+		seenURLs[urlString] = dest
+		// get the host from the url
+		parsedURL, err := url.Parse(urlString)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing url %s: %w", urlString, err)
+		}
+		host := parsedURL.Host
+		// add the url/dest pair to the manifestMap
+		if entries, ok := manifestMap[host]; !ok {
```
philandstuff, line 150: we should include the scheme too, because requests to http://example.com and https://example.com will use different connections

tempusfrangit: The only caveat is that this will not allow for the break on an empty line with a newline. I think you're right though, let me fix this to readfile.

cmd/multifile.go
tempusfrangit: Yeah Readfile is better. Will fix

cmd/multifile.go
tempusfrangit: Good callout.

cmd/multifile.go
tempusfrangit: Fixed in a subsequent patch.

cmd/multifile.go
tempusfrangit: Oh! We can't use ReadFile, since we need to rely on stdin as well, we can use readall though

cmd/multifile.go
tempusfrangit: ResolveReference takes a URL input. I think we should use a sprintf for url.Scheme and url.Host

tempusfrangit:
Comparison between 40 connections per host and 100:
40:
```
real	0m10.829s
user	0m40.103s
sys	1m10.355s
```

100:
```
real	0m8.373s
user	0m47.122s
sys	1m17.526s
```

491 files downloaded for each.

tempusfrangit:
I think I'm missing one final check in the code, I think we don't check if the destination directory exists so we might see failures without a mkdir -p equivalent.

PR #32, opened 2023-11-20
Replace custom roundtripper with retryablehttp

tempusfrangit:
Addresses #31

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -3,13 +3,13 @@ package client
 import (
 	"context"
 	"fmt"
-	"math"
 	"math/rand"
 	"net"
 	"net/http"
 	"net/url"
 	"time"
 
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/spf13/viper"
 
 	"github.com/replicate/pget/pkg/config"
@@ -19,12 +19,10 @@ import (
 )
 
 const (
-	retryDelayBaseline = 100 // in milliseconds
-	retrySleepJitter   = 500 // in milliseconds (will add 0-500 additional milliseconds)
+	retryMinWait     = 100 * time.Millisecond  // in milliseconds
+	retryMaxWait     = 3000 * time.Millisecond // in milliseconds, do not backoff further than 3 seconds
+	retrySleepJitter = 500                     // (will add 0-500 additional milliseconds), multiplied by time.Millisecond in backoffFunc
 
-	retryMaxBackoffTime = 3000 // in milliseconds, do not backoff further than 3 seconds
-	retryBackoffIncr    = 500  // in milliseconds, backoffFactor^retryNum * backoffIncr
-	retryBackoffFactor  = 2    // Base for POW()
 )
 
 var logger = logging.Logger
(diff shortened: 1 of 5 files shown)


pkg/client/client.go
```diff
@@ -109,20 +66,41 @@ func newClient(host string) *HTTPClient {
...
 	}
+
+	client := retryClient.StandardClient()
 	return &HTTPClient{Client: client, host: host}
 }
 
+// backoffFunc is a wrapper around retryablehttp.DefaultBackoff that allows for adding a random jitter to the backoff
+// we utilize the jitter to avoid thundering herd issues since we are running with significant numbers of concurrent
+// downloads.
+func backoffFunc(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration {
```
philandstuff, line 98: Did you see https://pkg.go.dev/github.com/hashicorp/go-retryablehttp ? For some reason upstream provides strict exponential or jittered linear backoff so I can see how you might have got here anyway.

philandstuff (approved):

pkg/client/client.go
tempusfrangit: I think we can just go with the linear jitter. I'll respin this PR to make that change as it lightens our carried code.

tempusfrangit: I wanted the Retry-After header support so here is a version that covers LinearBackoffWithJitter *and* Retry-After header.

pkg/client/client.go
```diff
@@ -19,12 +19,10 @@ import (
 )
 
 const (
-	retryDelayBaseline = 100 // in milliseconds
-	retrySleepJitter   = 500 // in milliseconds (will add 0-500 additional milliseconds)
-
-	retryMaxBackoffTime = 3000 // in milliseconds, do not backoff further than 3 seconds
-	retryBackoffIncr    = 500  // in milliseconds, backoffFactor^retryNum * backoffIncr
-	retryBackoffFactor  = 2    // Base for POW()
+	// These are boundings for the retryablehttp client and not aboslute values
```
philandstuff, line 22:
```suggestion
	// These are boundings for the retryablehttp client and not absolute values
```

pkg/client/client.go
```diff
@@ -109,20 +66,66 @@ func newClient(host string) *HTTPClient {
...
+func linearJitterRetryAfterBackoff(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration {
+	var retryAfter time.Duration
+
+	if shouldApplyRetryAfter(resp) {
+		retryAfter = evaluateRetryAfter(resp)
+	}
+
+	if retryAfter > 0 {
+		// If the Retry-After header is set, treat this as attempt 0 to get just the jitter
+		attemptNum = 0
```
philandstuff, line 105:
```suggestion
		retryablehttp.LinearJitterBackoff(retryAfter, retryAfter + jitter, 0, resp)
```
where `jitter` is `max - min`.

The current solution adds `retryAfter` to `min` which means we wait a bit more than we ought to.

philandstuff (approved):

PR #39, opened 2023-11-23
Fix test flake

philandstuff:
TestDownloadFilesFromHost was flaking.  This was because we were writing to dummyMode.args and dummyMode.timesCalled in an unsynchronised way from two different goroutines.  This meant that we could get lost updates.

I fixed this by adding a channel.  dummyMode.DownloadFile now sends its args to a channel; once you're done with your test input, you call dummyMode.Args() to read everything off the channel and put it into the slice as before.

### DIFF
--- a/cmd/multifile/multifile_test.go
+++ b/cmd/multifile/multifile_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


philandstuff:
I tested this worked by running `go test ./cmd/multifile -count=10000` a few times

tempusfrangit:
Included in #42

PR #40, opened 2023-11-23
Refactor client pool

philandstuff:
This is a refactoring of the client pooling:

* move away from global variables towards a dedicated ClientPool interface

* move away from global config (code in pkg/ shouldn't be making viper.Get* calls)

* make ClientPool responsible for acquiring and releasing clients, hiding it behind the standard Do() interface

* get rid of UserAgentTransport, now that we wrap Do() we can set User-agent there

* create client pools per scheme://host on demand, not up front.  This requires some funky work with a RWMutex so please review this carefully!

### DIFF
--- a/cmd/multifile/multifile.go
+++ b/cmd/multifile/multifile.go
@@ -12,7 +12,6 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/replicate/pget/pkg/cli"
-	"github.com/replicate/pget/pkg/client"
 	"github.com/replicate/pget/pkg/config"
 	"github.com/replicate/pget/pkg/download"
 	"github.com/replicate/pget/pkg/logging"
@@ -118,13 +117,6 @@ func multifileExecute(manifest manifest) error {
 		logging.Logger.Debug().Int("max_connections_per_host", perHostLimit).Msg("Config")
 	}
 
-	for schemeHost := range manifest {
-		err := handleConnectionPoolCreate(schemeHost)
-		if err != nil {
-			return err
-		}
-	}
-
 	// download each host's files in parallel
 	eg := initializeErrGroup()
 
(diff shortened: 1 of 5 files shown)


nickstenning:
As far as this goes the code looks good. 

I am overall a little dubious about the concept of pooling `http.Client`s. I know @tempusfrangit said there were potentially performance issues sharing a client over a single pool/transport, but that does stretch the limits of my imagination. In particularly, it's very hard to see how this code isn't ultimately doing exactly what `http.Transport` is doing internally, but with fewer tests ;)

Overall, I think my feedback is: if `pget` is a thing where we're going to do every last thing to optimize performance, then we should really have a performance test suite that tells us the performance impact of any mooted change (like this one, or like one to remove the concept of pooling `http.Client`s entirely).

tempusfrangit:
> As far as this goes the code looks good. 
> 
> 
> 
> I am overall a little dubious about the concept of pooling `http.Client`s. I know @tempusfrangit said there were potentially performance issues sharing a client over a single pool/transport, but that does stretch the limits of my imagination. In particularly, it's very hard to see how this code isn't ultimately doing exactly what `http.Transport` is doing internally, but with fewer tests ;)
> 
> 
> 
> Overall, I think my feedback is: if `pget` is a thing where we're going to do every last thing to optimize performance, then we should really have a performance test suite that tells us the performance impact of any mooted change (like this one, or like one to remove the concept of pooling `http.Client`s entirely).

The issue is that each client has ~3 mutexs for access to the transport. Sharing a client was significantly slower (maximum download of ~800MB/s-1GB/s across the same number of connections per host vs 1.5GB/s-4.2GB/s with isolated client per connection. 

We do need a benchmark suite.

tempusfrangit:
I can re-run the manual benchmark tomorrow on this change and start building a suite we can use for monitoring pget performance

philandstuff:
Going to merge as-is. We should definitely benchmark this against the status quo ante, and we should investigate if there's some way we're holding http.Client wrong too.

PR #41, opened 2023-11-24
Remove client pool

tempusfrangit:
ClientPool is not needed as long as we use HTTP/2 without keepalives
or HTTP/1.1. This is due to the multiplexing capabilities of HTTP/2 and
the inability to easily override the max streams value in the HTTP/2
client transport. We may want to consider implementing some deeper
changes eventually to allow us to use HTTP/2.

This change, due to removal of the client pool, now uses a single
shared connection and isolates just the reading of the resp.Body
to the goroutines.


tempusfrangit:
Verified the performance is within margins of error for release, main, this branch for pget multifile and pget single file

this version is *slightly* slower at 0.020s or so on avg in the single-file form, multifile is almost identical speed across ~39GB downloaded (within 10ms variation)

PR #49, opened 2023-11-27
Remove viper calls from client and download packages

philandstuff:
It was hard to test and maintain this code because it was making calls out to get configuration from global state.

This limits handling of configuration to cmd, pkg/cli, and pkg/config.

I removed the `ignore-retry-after` option because I couldn't find any usage of it and YAGNI.


pkg/download/buffer.go
```diff
@@ -37,12 +37,30 @@ func (t *Target) Basename() string {
...
-		Client: client,
+		Client:  client,
+		Options: opts,
 	}
 }
 
+func (m *BufferMode) maxChunks() int {
+	maxChunks := m.MaxChunks
+	if maxChunks == 0 {
+		return runtime.GOMAXPROCS(0) * 4
```
tempusfrangit, line 51: Can we move this to NumCPU() instead?

```suggestion
		return runtime.NumCPU(0) * 4
```

Or should we use the automaxprocs?

pkg/download/buffer.go
```diff
@@ -37,12 +37,30 @@ func (t *Target) Basename() string {
...
+	}
+	return maxChunks
+}
+
+func (m *BufferMode) minChunkSize() int64 {
+	minChunkSize := m.MinChunkSize
+	if minChunkSize == 0 {
+		return 16 * 1024 * 1024
```
tempusfrangit, line 59: Should this be a constant? Does this get optimized out in compilation?

No change needed here.

tempusfrangit (approved):

tempusfrangit (approved):

PR #51, opened 2023-11-27
Test and benchmark downloading

philandstuff:
This adds a test (based on an in-memory filesystem) and a few benchmarks (based on synthetic files created in a temp dir, and one real file in GCS).

Slow benchmarks are guarded by a `slow` build tag; they will only run if you add `--tags=slow` to your `go test` invocation.


tempusfrangit:
This looks good to me, lets merge and we'll iterate on it as needed. It's good to have a benchmark to start.

PR #53, opened 2023-11-28
Remove unneeded HEAD request

philandstuff:
In order to split a file up into chunks, we need to know the file size.  Right now we do that by starting with a HEAD request.

The way we do this leaks a connection. Because we don't read the response body in full (even though it's empty, being a HEAD request), net/http doesn't know that the response has finished using the connection.

I could fix this but.. it turns out we don't need the HEAD request at all!  We can optimistically make a GET request, with a Range header asking for the first MinChunkSize bytes.  Then we can inspect the Content-Range header in the response which will tell us the full file size, and we can divide up the rest of the file into chunks to pull down.

This slightly increases the median chunk size, but saves a network round trip.

Benchmarking on my macbook, old code:

go test ./pkg/download -bench=.
goos: darwin
goarch: arm64
pkg: github.com/replicate/pget/pkg/download
BenchmarkDownload10K-12       	    1771	    692691 ns/op
BenchmarkDownload10M-12       	     217	   5376755 ns/op
BenchmarkDownload100M-12      	      22	  50285470 ns/op
BenchmarkDownload1G-12        	       3	 353891486 ns/op
BenchmarkDownload10KH2-12     	    2460	    496717 ns/op
BenchmarkDownload10MH2-12     	     244	   4868315 ns/op
BenchmarkDownload100MH2-12    	      21	  50061675 ns/op
BenchmarkDownload1GH2-12      	       3	 348711125 ns/op
PASS
ok  	github.com/replicate/pget/pkg/download	30.270s

After this commit:

go test ./pkg/download -bench=.
goos: darwin
goarch: arm64
pkg: github.com/replicate/pget/pkg/download
BenchmarkDownload10K-12       	    2212	    492589 ns/op
BenchmarkDownload10M-12       	     236	   4988545 ns/op
BenchmarkDownload100M-12      	      22	  50301936 ns/op
BenchmarkDownload1G-12        	       3	 354903416 ns/op
BenchmarkDownload10KH2-12     	    3213	    403530 ns/op
BenchmarkDownload10MH2-12     	     285	   4091571 ns/op
BenchmarkDownload100MH2-12    	      24	  47704658 ns/op
BenchmarkDownload1GH2-12      	       3	 345798791 ns/op
PASS
ok  	github.com/replicate/pget/pkg/download	31.001s

As might be expected, files under the MinChunkSize are significantly faster; files over the MinChunkSize are about the same. Also we no longer leak connections.

### DIFF
--- a/pkg/download/buffer.go
+++ b/pkg/download/buffer.go
@@ -9,7 +9,9 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
+	"strconv"
 	"time"
 
 	"github.com/dustin/go-humanize"
@@ -21,6 +23,8 @@ import (
 
 const BufferModeName = "buffer"
 const defaultMinChunkSize = 16 * 1024 * 1024
+
+var contentRangeRegexp = regexp.MustCompile(`^bytes .*/([0-9]+)$`)
 
 type BufferMode struct {
 	Client *client.HTTPClient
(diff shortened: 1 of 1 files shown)


tempusfrangit (approved):

technillogue:
does this drop the trick of following the redirect only once?

philandstuff:
> does this drop the trick of following the redirect only once?

No, we still do this, see line 85

Issue #48, opened 2023-11-26
Pid file only created on root command


tempusfrangit:
The pid file is only created when running the root command. This should either be moved to main() or removed.

PR #56, opened 2023-11-28
Cleanup context

tempusfrangit:
Cleanup context (pass it through from root) and remove useless option related to max concurrent files. We do not need to leverage the errgroup to limit the number of downloads any longer. We're using MaxConnPerHost to do this.


tempusfrangit:
Addresses #48

PR #59, opened 2023-11-29
Refactor: introduce Strategy and Consumer interfaces

philandstuff:
Background work; I'm hoping consistent hashing will be another implementation of the Strategy interface.


pkg/pget.go
```diff
@@ -0,0 +1,52 @@
...
+	size := humanize.Bytes(uint64(fileSize))
+	throughput := humanize.Bytes(uint64(float64(fileSize) / float64(totalElapsed)))
+	logger.Debug().
+		Str("dest", dest).
+		Str("size", size).
+		Str("overall_throughput", throughput).
+		Str("download_elapsed", fmt.Sprintf("%.3fs", downloadElapsed.Seconds())).
+		Str("write_elapsed", fmt.Sprintf("%.3fs", writeElapsed.Seconds())).
+		Str("total_elapsed", fmt.Sprintf("%.3fs", totalElapsed.Seconds())).
+		Msg("Write Complete")
```
tempusfrangit, line 49: Let's make the logging in buffer.go be a more useful Debug (if anything) output, let's convert this one over to the .Info(), same with the writer interfaces. I think we should have interesting debug (if anything) there and consolidate the logging here for the info output. Maybe this is too much info now? 

Note: I switched the writeStartTime to be after the download complete 

First Option:
```suggestion
	downloadElapsed := time.Since(downloadStartTime)
	writeStartTime := time.Now()
	
	err = g.Consumer.Consume(buffer, dest)
	if err != nil {
		return fileSize, 0, fmt.Errorf("error writing file: %w", err)
	}
	writeElapsed := time.Since(writeStartTime)
	totalElapsed := time.Since(downloadStartTime)

	size := humanize.Bytes(uint64(fileSize))
	downloadThroughput := humanize.Bytes(uint64(float64(fileSize) / float64(downloadElapsed)))
	writeThroughput := humanize.Bytes(uint64(float64(fileSize) / float64(writeElapsed)))
 logger.Info().
		Str("dest", dest).
		Str("size", size).
		Str("write_throughput", writeThrougput).
		Str("download_throughput", downloadThroughput).
		Str("download_elapsed", fmt.Sprintf("%.3fs", downloadElapsed.Seconds())).
		Str("write_elapsed", fmt.Sprintf("%.3fs", writeElapsed.Seconds())).
		Str("total_elapsed", fmt.Sprintf("%.3fs", totalElapsed.Seconds())).
		Msg("Complete")
```
Alternate option:
keep super relevant (size, dest, total_elapsed, (download_throughput? since it was in the original output of pget) here in .Info() and add other things like write_elapsed and download_elapsed into an explicit .Debug() log call?

pkg/extract/tar.go
```diff
@@ -58,12 +55,8 @@ func ExtractTarFile(buffer *bytes.Buffer, destDir string, fileSize int64) error
 		}
 	}
 	elapsed := time.Since(startTime).Seconds()
-	size := humanize.Bytes(uint64(fileSize))
-	throughput := humanize.Bytes(uint64(float64(fileSize) / elapsed))
 	logger.Info().
-		Str("size", size).
 		Str("elapsed", fmt.Sprintf("%.3fs", elapsed)).
-		Str("throughput", throughput).
 		Msg("Extracted")
```
tempusfrangit, line 60: For now lets just drop this log. We can refactor this whole thing to be better (down the road) and add extra debugging such as "creating directory", "extracting file", "creating symlink", etc etc.

That is assuming we make the change below making the DownloadFile responsible for the final .Info() logging.

tempusfrangit (changes requested):

tempusfrangit:
This also breaks the -f flag (at least on OS X) 
`Error: error writing file: error writing file: open ./safety: file exists`

tempusfrangit (approved):

PR #60, opened 2023-11-30
Introduce null consumer and Hidden option to set consumer

tempusfrangit:
Implements a null consumer and a hidden option to set the consumer.


pkg/consumer/consumer.go
```diff
@@ -1,7 +1,23 @@
...
+	case "file":
+		return &FileWriter{}, nil
+	case "tar":
+		return &TarExtractor{}, nil
+	case "null":
+		return &NullWriter{}, nil
+	default:
+		return nil, fmt.Errorf("unknown consumer: %s", consumerName)
+	}
+}
```
tempusfrangit, line 23: I don't like this. Is there something better we can do? - I'd really like to have this functionality but I don't like the "GetXXX" style function. Not sure if there is something better we can do.

pkg/consumer/null.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+type NullWriter struct{}
+
+var _ Consumer = &NullWriter{}
+
+func (f *NullWriter) Consume(reader io.Reader, destPath string) error {
+	_, err := io.Copy(io.Discard, reader)
+	if err != nil {
+		return fmt.Errorf("error writing file: %w", err)
```
tempusfrangit, line 15:
```suggestion
		return fmt.Errorf("error discarding data: %w", err)
```

pkg/config/config.go
```diff
@@ -18,8 +18,8 @@ import (
 var (
 	ConnTimeout      time.Duration
 	Force            bool
+	Consumer         string
```
tempusfrangit, line 21:
```suggestion
	Consumer         string
	Force            bool
```

cmd/root/root.go
```diff
@@ -108,8 +108,13 @@ func rootExecute(ctx context.Context, urlString, dest string) error {
 		MinChunkSize: int64(minChunkSize),
 		Client:       clientOpts,
 	}
+	c, err := consumer.GetConsumer(viper.GetString(optname.Consumer))
```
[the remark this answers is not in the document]
tempusfrangit: Id like a common way to figure this out not "each command reimplementing. Like I said I don't like it as is. 

Maybe we just make the function in consumer (or move it to config) return the mapping of names to consumers? And cmd is responsible for doing the lookup? 

I'm trying to figure out the golang pattern here to turn a string into a struct type. 

I guess we could also add a flag per consumer and make them mutually exclusive?

tempusfrangit:
@philandstuff This can wait until consistent hashing lands and I'm happy to rebase it at that time. I don't want to cause too much more churn for you.

Let me know
1.  If this is better than before
2. If you want to land this before or after consistent hashing. I'll help with the rebase in either case.

PR #68, opened 2023-11-30
Remove files that snuck into tree

tempusfrangit:
Looks like my global git ignore failed at some point. This removes the files I inadvertently added.


philandstuff:
Oops, this was absolutely me and not you.

PR #69, opened 2023-11-30
Consistent hashing

philandstuff:
This introduces a consistent hashing mode for pget.

Based on configuration options for max concurrency, slice size, minimum chunk size, and the name of an SRV record to get cache hosts from, and a list of domains to redirect to the cache hosts, it will apply consistent hashing to requests to the allowlisted domains. That is, it will redirect a request to one of the cache hosts consistently, where the cache key is `http://example.com/path/file.ext#N`, where N is the slice number.

Files are sliced up into fixed-size slices which correspond to nginx's concept of cache slices. Each slice will always be sent to the same cache host.

Not done in this PR:

- getting the UX completely sorted
- retries of any sort (if a cache host is broken, we have no way to recover)


pkg/download/consistent_hashing.go
```diff
@@ -92,49 +101,61 @@ func (m *ConsistentHashingMode) Fetch(ctx context.Context, url string) (io.Reade
...
-		chunks = m.maxChunks()
-		chunkSize = fileSize / int64(chunks)
-	}
-	logger.Debug().Str("url", url).
-		Int64("size", fileSize).
-		Int("connections", chunks).
-		Int64("chunkSize", chunkSize).
-		Msg("Downloading")
+	totalSlices := fileSize / m.SliceSize
+	totalSlices++
```
philandstuff, line 111: this is a bug

philandstuff:
```suggestion
	if fileSize % m.SliceSize != 0 { totalSlices++ }
```

nickstenning (commented):
Only got a short way through this. Will come back and review the rest later.

cmd/root/root.go
```diff
@@ -102,10 +103,24 @@ func rootExecute(ctx context.Context, urlString, dest string) error {
 		MaxConcurrency: viper.GetInt(optname.Concurrency),
 		MinChunkSize:   int64(minChunkSize),
 		Client:         clientOpts,
+		Semaphore:      semaphore.NewWeighted(int64(viper.GetInt(optname.Concurrency))),
 	}
 	getter := pget.Getter{
 		Downloader: download.GetBufferMode(downloadOpts),
 	}
+	if srvName := viper.GetString(optname.CacheNodesSRVName); srvName != "" {
+		downloadOpts.SliceSize = 512 * 1024 * 1024 * 1024
```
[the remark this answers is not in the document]
philandstuff: resolved in e701b4d

pkg/download/consistent_hashing.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	trueURL := firstChunkResp.Request.URL.String()
+	if trueURL != url {
+		logger.Info().Str("url", url).Str("redirect_url", trueURL).Msg("Redirect")
+	}
+
+	fileSize, err := m.getFileSizeFromContentRange(firstChunkResp.Header.Get("Content-Range"))
+	if err != nil {
+		firstChunkResp.Body.Close()
```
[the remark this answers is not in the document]
philandstuff: I don't know what the right thing to do here is though - reading the body means pulling MinChunkSize bytes down the wire, so we need to weigh up the cost of doing that with the cost of dropping a connection

pkg/download/consistent_hashing.go
```diff
@@ -0,0 +1,210 @@
...
+	return &ConsistentHashingMode{
+		Client:  client,
+		Options: opts,
+	}, nil
+}
+
+func (m *ConsistentHashingMode) maxChunks() int {
+	maxChunks := m.MaxConcurrency
+	if maxChunks == 0 {
+		return runtime.NumCPU() * 4
```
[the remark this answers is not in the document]
tempusfrangit: Yeah, the 8x Nodes have a high GOMAXPROCS and high affinity so even NumCPU is probably too high.

tempusfrangit:
I do want to note that this branch broke the use of `--max-conn-per-host` should we emit a deprecation method vs "command fails"? 

I'm still reviewing the code.

Synthetic Benchmark against cache has shown:
* Equivalent performance to pre-CH pget using concurrency at the same number as --max-conn-per-host. No general regression

Benchmark was done in Coreweave from a lower resource (32Gi Ram, 16CPU pod) icelake pod to cache.
Using mlc-llama70b for muiltifile and SDXL for single file. Performance came out to near identical. Still working on a synthetic benchmark to show standard deviation when doing the downloads.

cmd/multifile/multifile.go
```diff
@@ -70,7 +70,6 @@ func GetCommand() *cobra.Command {
 		Example: multifileExamples,
 	}
 
-	cmd.PersistentFlags().Int(optname.MaxConnPerHost, 40, "Maximum number of (global) concurrent connections per host")
```
tempusfrangit, line 73: See my other comment- should this be a deprecation before pget errors saying "no such option"? We can don't need to use it even with a deprecation

philandstuff:
> I do want to note that this branch broke the use of --max-conn-per-host should we emit a deprecation method vs "command fails"?

I'm open to re-enabling it, but I agree with @nickstenning that this shouldn't be our primary source of concurrency control, more a backstop.

pkg/cli/common.go
```diff
@@ -44,3 +48,48 @@ func EnsureDestinationNotExist(dest string) error {
...
+		if err != nil {
+			return nil, err
+		}
+		hostname := strings.TrimSuffix(srv.Target, ".")
+		if srv.Port != 80 {
+			hostname = fmt.Sprintf("%s:%d", hostname, srv.Port)
+		}
+		output[cacheIndex] = hostname
+	}
+	return output, nil
```
tempusfrangit, line 86: This doesn't need to be changed to land the code:

We could avoid the second pass through `srvs`, but i'm not sure if that would improve the readability of the code. I'm making this comment because it took me longer than I'd like to admit to tease out the dual loops.

pkg/client/client.go
```diff
@@ -59,8 +58,9 @@ func NewHTTPClient(opts Options) *HTTPClient {
 		ExpectContinueTimeout: 1 * time.Second,
 		DisableKeepAlives:     disableKeepAlives,
 	}
-	transport.MaxConnsPerHost = opts.MaxConnPerHost
-	transport.MaxIdleConnsPerHost = opts.MaxConnPerHost
+
+	// FIXME should we set this to something larger than the default of 2?
+	// transport.MaxIdleConnsPerHost = opts.MaxConnPerHost
```
tempusfrangit, line 63: Let's make sure to update this before we land the code. Easiest is to set it == concurrency. We could just do Concurrency+5 for super easy math or some such.

pkg/download/buffer.go
```diff
@@ -13,11 +13,13 @@ import (
...
 	"golang.org/x/sync/errgroup"
 
+	"github.com/dustin/go-humanize"
+
 	"github.com/replicate/pget/pkg/client"
 	"github.com/replicate/pget/pkg/logging"
 )
 
-const defaultMinChunkSize = 16 * 1024 * 1024
+const defaultMinChunkSize = 16 * humanize.MiByte
```
tempusfrangit, line 22: Nice!

pkg/download/consistent_hashing.go
```diff
@@ -0,0 +1,263 @@
...
+	shouldContinue := false
+	for _, host := range m.DomainsToCache {
+		if host == parsed.Host {
+			shouldContinue = true
+			break
+		}
+	}
+	if !shouldContinue {
+		return nil, -1, fmt.Errorf("ConsistentHashingMode not implemented for domains outside DomainsToCache")
+	}
```
tempusfrangit, line 83: We could make this work for *anything* in reality with a little thought to the cache nginx configs. I am not sure we want to though.

pkg/download/consistent_hashing.go
```diff
@@ -0,0 +1,263 @@
...
+	req, err := http.NewRequestWithContext(ctx, "GET", urlString, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to download %s: %w", req.URL.String(), err)
+	}
+	err = m.consistentHashIfNeeded(req, start, end)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", start, end))
```
tempusfrangit, line 204: NIT: I'd love to get "Range" into a constant just so future things can't toyp it, same with bytes=.

Not worth changing here.

pkg/download/consistent_hashing.go
```diff
@@ -0,0 +1,263 @@
...
+	return resp, nil
+}
+
+func (m *ConsistentHashingMode) consistentHashIfNeeded(req *http.Request, start int64, end int64) error {
+	logger := logging.GetLogger()
+	for _, host := range m.DomainsToCache {
+		if host == req.URL.Host {
+			if start/m.SliceSize != end/m.SliceSize {
+				return fmt.Errorf("Can't make a range request across a slice boundary: %d-%d straddles a slice boundary (slice size is %d)", start, end, m.SliceSize)
+			}
```
tempusfrangit, line 225: Is this mostly suspenders and belt? We should be able to avoid this scenario in all cases with something like:
minChunkSize > SliceSize -> minChunkSize == SliceSize -> chunkSize = SliceSize

SliceSize > minChunkSize > SliceSize/2 -> Round up or RoundDown? Preference roundDown and know we'll just use 2 chunks per slice always.

SliceSize % minChunkSize != 0, and does not fall into the above buckets -- we already handle this by having a smaller final chunk


To be clear, this code is fine as is. Thinking about how to make it more robust long term.

pkg/download/consistent_hashing_test.go
```diff
@@ -0,0 +1,186 @@
...
+	{ // if minChunkSize > sliceSize, sliceSize overrides it
+		concurrency:    8,
+		sliceSize:      3,
+		numCacheHosts:  8,
+		minChunkSize:   24,
+		expectedOutput: "1117770002226667",
+	},
+}
+
+func TestConsistentHashing(t *testing.T) {
```
tempusfrangit, line 153: This is a fun test.

tempusfrangit (commented):
Only thing I'd like fixed is the MaxIdleConnsPerHost set to a better value and MaxIdleConns==0.

pkg/cli/common.go
philandstuff: yeah it's a go thing. In another language I might have a single function to find the highest cache index but Go really wants to make it explicit when you're looping through a slice so you don't accidentally write O(n^2) code when you think you're writing O(n) code.

I could add comments to make it clearer what the purpose of each loop is?

pkg/download/consistent_hashing.go
philandstuff: yeah I'd like this to fall back to BufferMode in future

pkg/download/consistent_hashing.go
philandstuff: yeah this is a "can't happen" guard

PR #74, opened 2023-12-06
Fix bad range req min chunksize

### LINKED ISSUE #73: BUG: downloads that do not use the standard chunksize in 0.2.0 and above result in broken files
0.2.0+ of pget results in broken files when/if a download is performed due to a miscalculation in the range-request.

tempusfrangit:
* Implement test to expose byte-range-request bug
* Implement fix to incorrect range-requests when maxChunks < fileSize / minChunkSize

Closes #73

### DIFF
--- a/pkg/download/buffer_unit_test.go
+++ b/pkg/download/buffer_unit_test.go
@@ -2,15 +2,16 @@ package download
 
 import (
 	"context"
-	"github.com/dustin/go-humanize"
-	"github.com/rs/zerolog"
-	"github.com/stretchr/testify/assert"
 	"math/rand"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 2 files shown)


pkg/download/buffer_unit_test.go
```diff
@@ -64,6 +64,11 @@ func TestFileToBufferChunkCountExceedsMaxChunks(t *testing.T) {
 			minChunkSize: contentSize + 1,
 			maxChunks:    1,
 		},
+		{
+			name:         "minChunkSize equal to contentSize",
+			minChunkSize: contentSize,
+			maxChunks:    1,
+		},
```
tempusfrangit, line 71: This is an additional edgecase that resulted in divide by zero errors.

PR #97, opened 2023-12-10
Cleanup

tempusfrangit:
* Eliminate global host resolution map - generally less fragile avoiding a global causing behavior changes.
  * Resolution overrides are now passed down in the client options.
 
* Make the same fix made to consistent-hashing and move the .doRequest to within the go routine preventing blocking the main thread for no good reason.


philandstuff (approved):
I'll push a commit for that one change and then merge

PR #99, opened 2023-12-10
Implement consistent-hashing for multifile

tempusfrangit:
Multifile MUST support consistent-hashing mode. This is a copy/paste from the rootCMD override of the getter.Downloader; we should DRY this eventually.

Closes: #98


philandstuff:
This does nothing to ensure overall concurrency control; we need to do the errgroup limit thing or the semaphore thing. I'll pick this PR up and implement those pieces.

tempusfrangit:
Wouldn't the semaphore provide concurrency management? We use a single unified client and a single getter. 

I'm happy to add the regroup limiting code (and prefer it based on our convo last week).

PR #104, opened 2023-12-11
Refactor multifile

philandstuff:
I want to implement better concurrency control via a shared semaphore or limited errGroup; but to get started I wanted to move this logic into the pget package.


tempusfrangit:
LGTM!

PR #108, opened 2023-12-12
Implement consistent-hash fallback for files and chunks

tempusfrangit:
This commit implements consistent hash fallback behavior when a file's first request fails or a given chunk fails. This is handled in the retryablehttp client's `RetryPolicy` function. In any case that a fall-back is desired, the `RetryPolicy` will return a `false` (meaning do not retry) and a special `ErrStrategyFallBack` that the `Strategy` can act upon.

This commit exposes DoRequest as part of the Strategy Interface to handle the fallback for a single chunk.

`RetryPolicy`'s additional logic above (and beyond) the `DefaultRetryPolicy` will only trigger when using consistent-hash strategy.

Closes: #107


pkg/client/client.go
```diff
@@ -76,14 +80,71 @@ func NewHTTPClient(opts Options) *HTTPClient {
...
-		CheckRetry:   retryablehttp.DefaultRetryPolicy,
+		CheckRetry:   RetryPolicy,
 		Backoff:      linearJitterRetryAfterBackoff,
 	}
 
 	client := retryClient.StandardClient()
 	return &HTTPClient{Client: client}
 }
 
+// RetryPolicy wraps retryablehttp.ErrorPropagatedRetryPolicy and included additional logic:
```
tempusfrangit, line 91:
```suggestion
// RetryPolicy wraps retryablehttp.DefaultRetryPolicy and included additional logic:
```

pkg/download/consistent_hashing.go
```diff
@@ -37,9 +41,10 @@ func GetConsistentHashingMode(opts Options) (Strategy, error) {
 	fallbackStrategy.Client = client
 
 	return &ConsistentHashingMode{
-		Client:           client,
-		Options:          opts,
-		FallbackStrategy: fallbackStrategy,
+		Client:               client,
+		Options:              opts,
+		FallbackStrategy:     fallbackStrategy,
+		FallbackStrategyType: reflect.TypeOf(fallbackStrategy),
```
tempusfrangit, line 47:
```suggestion
```

pkg/download/consistent_hashing.go
```diff
@@ -3,26 +3,30 @@ package download
...
 	"github.com/replicate/pget/pkg/logging"
 )
 
 type ConsistentHashingMode struct {
 	Client *client.HTTPClient
 	Options
 	// TODO: allow this to be configured and not just "BufferMode"
-	FallbackStrategy Strategy
+	FallbackStrategy     Strategy
+	FallbackStrategyType any
```
tempusfrangit, line 29:
```suggestion
```

pkg/client/client.go
```diff
@@ -76,14 +80,71 @@ func NewHTTPClient(opts Options) *HTTPClient {
...
+	// extra processing logic in these cases every time
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+
+	// While type assertions are not ideal, alternatives are limited to adding custom data in the request
+	// or in the context. The context clearly isolates this data.
+	consistentHashing, ok := ctx.Value(config.ConsistentHashingStrategyKey).(bool)
+	if ok && consistentHashing {
+		if err != nil && fallbackError(err) {
```
philandstuff, line 106:
```suggestion
		if fallbackError(err) {
```

pkg/client/client.go
```diff
@@ -76,14 +80,71 @@ func NewHTTPClient(opts Options) *HTTPClient {
...
+	// Wrap the standard retry policy
+	return retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+}
+
+// fallbackError returns true if the error is an error we should fall back to the next strategy.
+// fallback errors are not retryable errors that indicate fundamental problems with the cache-server
+// or networking to the cache server. These errors include connection timeouts, connection refused, dns
+// lookup errors, etc.
+func fallbackError(err error) bool {
+	if err != nil {
```
philandstuff, line 123: I'd rather `if err == nil { return false }` here

pkg/client/client.go
```diff
@@ -76,14 +80,71 @@ func NewHTTPClient(opts Options) *HTTPClient {
...
+	// do not retry on context.Canceled or context.DeadlineExceeded, this is a fast-fail even though
+	// the retryablehttp.ErrorPropagatedRetryPolicy will also return false for these errors. We can avoid
+	// extra processing logic in these cases every time
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+
+	// While type assertions are not ideal, alternatives are limited to adding custom data in the request
+	// or in the context. The context clearly isolates this data.
+	consistentHashing, ok := ctx.Value(config.ConsistentHashingStrategyKey).(bool)
```
philandstuff, line 104: out of interest, why make this consistent-hashing specific? if regular buffer mode hits a similar unrecoverable error, we could return ErrStrategyFallback and buffer mode just aborts. This is a change to current behaviour but if we're hitting 502 or 503 we're unlikely to retry our way out of that problem anyway.

(We can address this in another PR)

pkg/client/client_test.go
```diff
@@ -15,3 +21,136 @@ func TestGetSchemeHostKey(t *testing.T) {
...
+			resp:           &http.Response{},
+			err:            &net.OpError{Op: "dial"},
+			expectedResult: false,
+			expectedError:  client.ErrStrategyFallback,
+		},
+		{
+			name:           "net.OpErr: read",
+			ctx:            chCtx,
+			resp:           &http.Response{},
+			err:            &net.OpError{Op: "read"},
```
philandstuff, line 61: can we have a test for a non-{dial,read} op here?

pkg/client/client_test.go
```diff
@@ -15,3 +21,136 @@ func TestGetSchemeHostKey(t *testing.T) {
...
+			resp:           &http.Response{},
+			err:            &net.DNSError{IsTimeout: true},
+			expectedResult: false,
+			expectedError:  client.ErrStrategyFallback,
+		},
+		{
+			name:           "net.DNSErr: IsTemporary",
+			ctx:            chCtx,
+			resp:           &http.Response{},
+			err:            &net.DNSError{IsTemporary: true},
```
philandstuff, line 77: Can we have a test for a DNSError that we think is recoverable?

pkg/client/client_test.go
```diff
@@ -15,3 +21,136 @@ func TestGetSchemeHostKey(t *testing.T) {
...
+			name:           "Status OK",
+			ctx:            chCtx,
+			resp:           &http.Response{StatusCode: http.StatusOK},
+			err:            urlError,
+			expectedResult: false,
+		},
+		{
+			name:           "Status Service Unavailable",
+			ctx:            chCtx,
+			resp:           &http.Response{StatusCode: http.StatusBadGateway},
```
philandstuff, line 113:
```suggestion
			resp:           &http.Response{StatusCode: http.StatusServiceUnavailable},
```
copypaste fail

pkg/client/client_test.go
```diff
@@ -15,3 +21,136 @@ func TestGetSchemeHostKey(t *testing.T) {
...
+			expectedResult: false,
+			expectedError:  client.ErrStrategyFallback,
+		},
+		{
+			name:           "Recoverable Error",
+			ctx:            chCtx,
+			resp:           &http.Response{StatusCode: http.StatusTooManyRequests},
```
philandstuff, line 120:
```suggestion
			resp:           &http.Response{StatusCode: http.StatusOK},
```
otherwise this is confusing with the next case

pkg/download/strategy.go
```diff
@@ -2,9 +2,27 @@ package download
...
+	// DoRequest sends an HTTP GET request with a specified range of bytes to the given URL using the provided context.
+	// It returns the HTTP response and any error encountered during the request. It is intended that Fetch calls DoRequest
+	// and that each chunk is downloaded with a call to DoRequest. DoRequest is exposed so that consistent-hashing can
+	// utilize any strategy as a fall-back for chunk downloading.
+	//
+	// If the request fails to download or execute, an error is returned.
+	//
+	// The start and end parameters specify the byte range to request.
+	// The trueURL parameter is the actual URL after any redirects.
+	DoRequest(ctx context.Context, start, end int64, url string) (*http.Response, error)
```
philandstuff, line 27: We're introducing this as a test seam; it shouldn't be on this interface but on a test-specific interface

philandstuff, line 27: I don't like adding this to the Strategy interface. I'll think about if we can make it nicer.

pkg/client/client_test.go
philandstuff: Also, do we really think `IsTemporary` is an error that shouldn't be retried?

pkg/download/strategy.go
philandstuff: nope, can't think how to make it nicer

philandstuff (approved):

pkg/download/strategy.go
tempusfrangit: That was where I landed. Tried a number of different ways and just finally said *sigh* I guess this is it. 

I almost added a fallback structure that did the lifting but it was about the same and didn't guarantee any future strategy would define the needed interface.

PR #111, opened 2023-12-12
consistent-hashing using hashstructure

philandstuff:
This adopts mitchellh/hashstructure, so that we can add keys to the hash at a later date if we decide to.

Things we might consider adding in future:

* a retry field, so if a node fails we distribute retries to other nodes in the cluster in a consistent way
* a random field, for hot downloads, to avoid overwhelming a single cache host

We don't need these now but we're just reserving the ability to implement it later.


nickstenning (approved):
LGTM!

tempusfrangit (approved):
LGTM

PR #115, opened 2023-12-12
Better concurrency

philandstuff:
We want to enforce a global maximum concurrency limit:

- no more than N requests in flight, total
- as soon as one request finishes, another may begin
- the same concurrency limit is shared between BufferMode and ConsistentHashingMode for fallback reasons

This is aiming towards implementing that.


pkg/pget.go
```diff
@@ -39,26 +39,27 @@ func (g *Getter) DownloadFile(ctx context.Context, url string, dest string) (int
...
+		Str("url", url).
 		Str("size", size).
-		Str("download_throughput", fmt.Sprintf("%s/s", downloadThroughput)).
-		Str("download_elapsed", fmt.Sprintf("%.3fs", downloadElapsed.Seconds())).
-		Str("write_throughput", fmt.Sprintf("%s/s", writeThroughput)).
-		Str("write_elapsed", fmt.Sprintf("%.3fs", writeElapsed.Seconds())).
+		// Str("download_throughput", fmt.Sprintf("%s/s", downloadThroughput)).
+		// Str("download_elapsed", fmt.Sprintf("%.3fs", downloadElapsed.Seconds())).
+		// Str("write_throughput", fmt.Sprintf("%s/s", writeThroughput)).
+		// Str("write_elapsed", fmt.Sprintf("%.3fs", writeElapsed.Seconds())).
```
tempusfrangit, line 62: We lost the overall throughput and such for the single file mode. Maybe we can push that data back up to cmd.root to look the same as multifile with `[ metrics ]` as the .Msg()?

pkg/pget.go
```diff
@@ -70,6 +71,8 @@ func (g *Getter) DownloadFiles(ctx context.Context, manifest Manifest) (int64, t
 	}
 
 	errGroup, ctx := errgroup.WithContext(ctx)
+	// TODO: un-hardcode this
+	errGroup.SetLimit(20)
```
tempusfrangit, line 75: I'd really like to make this at the very least a ENV configurable if not a hidden cli flag.

pkg/download/consistent_hashing.go
```diff
@@ -154,80 +169,77 @@ func (m *ConsistentHashingMode) Fetch(ctx context.Context, urlString string) (io
...
-		}
-		if slice == int(totalSlices)-1 {
-			sliceSize = (fileSize-1)%m.SliceSize + 1
-		}
-		if sliceSize/numChunks < m.minChunkSize() {
-			// reset numChunks to respect minChunkSize
-			numChunks = sliceSize / m.minChunkSize()
-			// although we must always have at least one chunk
+	m.queue.submit(func() {
+		defer close(readersCh)
```
tempusfrangit, line 182: Same as the other comment -- this is a lot of code in the closure. I'd like it to be isolated to external function eventually.

pkg/download/consistent_hashing.go
```diff
@@ -100,46 +102,59 @@ func (m *ConsistentHashingMode) Fetch(ctx context.Context, urlString string) (io
 		return m.FallbackStrategy.Fetch(ctx, urlString)
 	}
 
-	firstChunkResp, err := m.DoRequest(ctx, 0, m.minChunkSize()-1, urlString)
-	if err != nil {
+	br := newBufferedReader(m.minChunkSize())
+	firstReqResultCh := make(chan firstReqResult)
+	m.queue.submit(func() {
+		m.sem.Go(func() error {
```
tempusfrangit, line 108: Same as the other comment(s). I'd like these extracted to be less code, it's a lot of code in the closure. But this can be addressed later.

pkg/download/consistent_hashing.go
```diff
@@ -100,46 +102,59 @@ func (m *ConsistentHashingMode) Fetch(ctx context.Context, urlString string) (io
...
+				return err
+			}
+			firstReqResultCh <- firstReqResult{fileSize: fileSize, trueURL: trueURL}
+
+			return br.downloadBody(firstChunkResp)
+		})
+	})
+	firstReqResult, ok := <-firstReqResultCh
+	if !ok {
+		panic("shouldn't happen")
```
tempusfrangit, line 135: This is a good use of panic but I'd like the message to be descriptive in case it's hit in the real world: "shouldn't happen" is going to require searching the code to find where it came from (Even though I know today).

pkg/download/chan_multi_reader.go
```diff
@@ -0,0 +1,40 @@
+package download
+
+import "io"
+
+type chanMultiReader struct {
```
tempusfrangit, line 5: should we have a `var _ io.Reader = chanMultiReader" here?

tempusfrangit (approved):
This looks good as is, with one exception:

The loss of the logging output for the single-file mode. Once we restore that I think it's good. I don't think we need to block merging on the log fix.

pkg/pget.go
philandstuff: (we had this conversation IRL but just jotting down my position for the peanut gallery)

I think we need to be very specific about what we're trying to measure and why. Given we download individual files in parallel and during download we might have arbitrary blocking on the semaphore, what measurement do we want to make to indicate how fast an individual file was downloaded?

Similar questions apply to the writeThroughput. Given we now interleave writing with downloading, what specifically can/should we measure to show write throughput?

I do understand that it'd be good to identify where any bottlenecks are but I don't think there's a meaningful way we can say how fast a given file was downloaded. We can maybe do something for downloading chunks, but we wouldn't want a log line for every single chunk. Maybe we collect per-chunk download statistics (mean, P05/P50/P95 size/duration/throughput) and publish them? Possibly by upstream host?

pkg/download/buffer.go
```diff
@@ -93,51 +122,52 @@ func (m *BufferMode) fileToBuffer(ctx context.Context, url string) (*bytes.Buffe
...
-		Int64("chunkSize", chunkSize).
-		Msg("Downloading")
-
-	errGroup, ctx := errgroup.WithContext(ctx)
-	errGroup.Go(func() error {
-		return m.downloadChunk(firstChunkResp, data[0:m.minChunkSize()])
-	})
-
-	startOffset := m.minChunkSize()
+	m.queue.submit(func() {
```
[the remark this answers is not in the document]
philandstuff: I did try doing this but there's quite a lot of data that the closure closes over that would have to be passed in some other way and it made things more complicated rather than simpler.

pkg/download/consistent_hashing.go
philandstuff: agreed. also note to self: I did panic here but return error in buffer.go, it should be consistent.

pkg/pget.go
philandstuff: I'd also say that `download_elapsed` is broken in v0.4.1 already.

Issue #109, opened 2023-12-12
Move to go-releaser


tempusfrangit:
@anotherjesse we should totally move to go-releaser :)

PR #116, opened 2023-12-12
Migrate to GoReleaser

tempusfrangit:
Use goreleaser insead of `ncipollo/release-action@v1`

Closes: #109


tempusfrangit:
This PR does not retrofit the makefile to use goreleaser. That is intended to be a future PR.

PR #117, opened 2023-12-14
Goreleaser enhancements

tempusfrangit:
make now generally leans on goreleaser instead of a separate build process

version string has much more interesting data in it.

better testing

.gitignore ignores  `dist`

CI for goreleaser config


tempusfrangit:
hmm. Ci for goreleaser config isn't running.

tempusfrangit:
@philandstuff My biggest concern here is the use of goreleaser as part of generalized "make" if you don't see that as an issue I'm content to move forward and get the build processes consistent.

Makefile
```diff
@@ -77,3 +58,24 @@ format:
...
+.PHONY: check-goreleaser
+check-goreleaser:
+	@command -v goreleaser >/dev/null 2>&1 || { echo >&2 "goreleaser is required but not installed. Aborting. Run 'make install-goreleaser' to install"; exit 1; }
+
+.PHONY: install-goreleaser
+install-goreleaser:
+	@command -v goreleaser >/dev/null 2>&1 || { \
+		echo >&2 "goreleaser is required but not installed. Installing..."; \
+		curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh; \
+	}
```
mattt, line 71: Would it make sense to install goreleaser as part of tools.go?

tempusfrangit: Not sure if we can do that. I'll look into it. But I think goreleaser doesn't work like that.  Totally 100% worth the exploration though.

tempusfrangit: Also that url doesn't work :P

tempusfrangit: HaH! We can! `go install github.com/goreleaser/goreleaser@latest` sweet!

PR #119, opened 2023-12-14
Consistent hashing: consistently retry another host

philandstuff:
This implements a retry policy to consistently retry another host within the SRV set if we get a failure talking to the original host.

We still use the original fallback strategy if this retry fails. (Should we?)

### DIFF
--- a/pkg/consistent/consistent.go
+++ b/pkg/consistent/consistent.go
@@ -1,0 +1,45 @@
+// Package consistent implements consistent hashing for cache nodes.
+package consistent
+
+import (
+	"fmt"
+	"slices"
+
+	"github.com/dgryski/go-jump"
+	"github.com/mitchellh/hashstructure/v2"
+)
+
+type cacheKey struct {
+	Key     any
+	Attempt int
+}
+
+// HashBucket returns a bucket from [0,buckets). If you want to implement a
+// retry, you can pass previousBuckets, which indicates buckets which must be
+// avoided in the output. HashBucket will modify the previousBuckets slice by
+// sorting it.
+func HashBucket(key any, buckets int, previousBuckets ...int) (int, error) {
+	if len(previousBuckets) >= buckets {
+		return -1, fmt.Errorf("No more buckets left: %d buckets available but %d already attempted", buckets, previousBuckets)
+	}
+	// we set IgnoreZeroValue so that we can add fields to the hash key
+	// later without breaking things.
+	// note that it's not safe to share a HashOptions so we create a fresh one each time.
+	hashopts := &hashstructure.HashOptions{IgnoreZeroValue: true}
+	hash, err := hashstructure.Hash(cacheKey{Key: key, Attempt: len(previousBuckets)}, hashstructure.FormatV2, hashopts)
+	if err != nil {
+		return -1, fmt.Errorf("error calculating hash of key: %w", err)
+	}
+
+	// jump is an implementation of Google's Jump Consistent Hash.
+	//
+	// See http://arxiv.org/abs/1406.2294 for details.
+	bucket := int(jump.Hash(hash, buckets-len(previousBuckets)))
+	slices.Sort(previousBuckets)
+	for _, prev := range previousBuckets {
+		if bucket >= prev {
+			bucket++
+		}
+	}
+	return bucket, nil
+}
--- a/pkg/download/consistent_hashing.go
+++ b/pkg/download/consistent_hashing.go
@@ -9,12 +9,11 @@ import (
 	"net/url"
 	"strconv"
 
-	jump "github.com/dgryski/go-jump"
-	"github.com/mitchellh/hashstructure/v2"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/replicate/pget/pkg/client"
 	"github.com/replicate/pget/pkg/config"
+	"github.com/replicate/pget/pkg/consistent"
 	"github.com/replicate/pget/pkg/logging"
 )
 
@@ -32,7 +31,6 @@ type ConsistentHashingMode struct {
 type CacheKey struct {
 	URL   *url.URL `hash:"string"`
 	Slice int64
-	Retry bool
 }
 
 func GetConsistentHashingMode(opts Options) (*ConsistentHashingMode, error) {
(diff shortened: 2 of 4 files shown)


pkg/download/consistent_hashing.go
```diff
@@ -32,6 +32,7 @@ type ConsistentHashingMode struct {
 type CacheKey struct {
 	URL   *url.URL `hash:"string"`
 	Slice int64
+	Retry bool
```
nickstenning, line 35: Maybe I'm overcomplicating things but perhaps this should be an integer `RetryCount` to allow retries to hit more than one alternative node? We can cap it at one retry elsewhere in the code if we're worried about disturbing the cache layout, but it seems that this would give us a little more flexibility in future.

pkg/download/consistent_hashing.go
```diff
@@ -262,7 +263,28 @@ func (m *ConsistentHashingMode) DoRequest(ctx context.Context, start, end int64,
 
 	resp, err := m.Client.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("error executing request for %s: %w", req.URL.String(), err)
+		if errors.Is(err, client.ErrStrategyFallback) {
```
nickstenning, line 266: With a big caveat that I don't know the rest of this code very well, this looks like a surprising amount of code to implement a retry. Do we need to extract a function or a method here so that the code for making the initial request and the code to make the retry request only differ in a passed-in boolean `retry` (or an integer `retryCount` if you agree with my earlier suggestion)?

pkg/download/consistent_hashing.go
```diff
@@ -285,23 +307,68 @@ func (m *ConsistentHashingMode) consistentHashIfNeeded(req *http.Request, start
 			// later without breaking things.
 			// note that it's not safe to share a HashOptions so we create a fresh one each time.
 			hashopts := &hashstructure.HashOptions{IgnoreZeroValue: true}
-			hash, err := hashstructure.Hash(key, hashstructure.FormatV2, hashopts)
+			var hash uint64
+			hash, err = hashstructure.Hash(key, hashstructure.FormatV2, hashopts)
 			if err != nil {
-				return fmt.Errorf("error calculating hash of key")
+				err = fmt.Errorf("error calculating hash of key: %w", err)
+				return
```
nickstenning, line 314: Maybe we should add a linter rule for this, but I'm generally not a big fan of implicit `return` in Go. Is it reasonable to make this explicit with a value for `cachePodIndex`?

pkg/download/consistent_hashing.go
```diff
@@ -285,23 +307,68 @@ func (m *ConsistentHashingMode) consistentHashIfNeeded(req *http.Request, start
...
+			}
+
+			logger.Debug().Uint64("hash_sum", hash).Int("len_cache_hosts", len(m.CacheHosts)).Msg("consistent hashing")
+
+			// jump is an implementation of Google's Jump Consistent Hash.
+			//
+			// See http://arxiv.org/abs/1406.2294 for details.
+
+			// we advance around the ring by somewhere between 1 and n-1
+			displacement := int(jump.Hash(hash, len(m.CacheHosts)-1)) + 1
```
nickstenning, line 364: As with my comment above, this isn't necessarily a blocker right now but I'm a little bit uneasy that we have the actual consistent hash implementation duplicated in multiple places. It feels this (and the hashstructure/key calculations) should probably only be defined once and then if necessary used by glue code that uses the HTTP client.

pkg/download/consistent_hashing.go
tempusfrangit: Agreed. I don't think we will go beyond 2 retries unless the cache footprint is very wide but that exposes some provider specific issues.  This keeps flexibility without needing to cause a complete cache shuffle to change the key for a retry increase.

tempusfrangit (approved):
Approve with the caveat that I have not tested the code directly in the validation environment. It looks sane. I largely agree with @nickstenning 's comments.

pkg/download/consistent_hashing.go
philandstuff: done. I called it `Attempt`.

pkg/download/consistent_hashing.go
philandstuff: I've done a refactor and removed this duplication.

pkg/download/consistent_hashing.go
philandstuff: done

pkg/download/consistent_hashing.go
philandstuff: done

pkg/consistent/consistent.go
```diff
@@ -0,0 +1,45 @@
...
+	// See http://arxiv.org/abs/1406.2294 for details.
+	bucket := int(jump.Hash(hash, buckets-len(previousBuckets)))
+	slices.Sort(previousBuckets)
+	for _, prev := range previousBuckets {
+		if bucket >= prev {
+			bucket++
+		}
+	}
+	return bucket, nil
+}
```
tempusfrangit, line 45: This is so nice to have it extracted!

pkg/download/consistent_hashing.go
```diff
@@ -262,7 +261,28 @@ func (m *ConsistentHashingMode) DoRequest(ctx context.Context, start, end int64,
...
+				// return origErr so that we can use our regular fallback strategy
+				return nil, origErr
+			}
+			req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", start, end))
+			logger.Debug().Str("url", urlString).Str("munged_url", req.URL.String()).Str("host", req.Host).Int64("start", start).Int64("end", end).Msg("retry request")
+
+			resp, err = m.Client.Do(req)
+			if err != nil {
+				// return origErr so that we can use our regular fallback strategy
+				return nil, origErr
```
tempusfrangit, line 281: I think we're going to want this for the foreseeable future. As reliability is the real goal, I want to be durable for "cache disappears" but we still serve inference.

That said, I hear the earlier comments about "do we need this".

tempusfrangit (approved):
LGTM! I still have not tested in the validation environment. But I am willing to land this and iterate on further releases as we need for minor adjustments/critical oversights.

pkg/consistent/consistent.go
```diff
@@ -0,0 +1,45 @@
...
+type cacheKey struct {
+	Key     any
+	Attempt int
+}
+
+// HashBucket returns a bucket from [0,buckets). If you want to implement a
+// retry, you can pass previousBuckets, which indicates buckets which must be
+// avoided in the output. HashBucket will modify the previousBuckets slice by
+// sorting it.
+func HashBucket(key any, buckets int, previousBuckets ...int) (int, error) {
```
nickstenning, line 21: This is really nice. I like the derived/wrapped key a lot.

nickstenning (approved):

PR #121, opened 2023-12-15
Implement pidfile Flock serialization

tempusfrangit:


cmd/root/root.go
```diff
@@ -77,6 +96,18 @@ func GetCommand() *cobra.Command {
 	return cmd
 }
 
+// defaultPidFilePath returns the default path for the PID file. Notably modern OS X variants
+// do not have /var/run/
```
tempusfrangit, line 100:
```suggestion
// do not have /var/run or /run
```

pkg/cli/pid.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+type ProcessSerialization interface {
+	Acquire() error
+	Release() error
+}
```
tempusfrangit, line 12: This is useless... delete
```suggestion
```

cmd/root/root.go
```diff
@@ -50,6 +50,25 @@ func GetCommand() *cobra.Command {
...
+			pidFilePath := viper.GetString(config.OptPIDFile)
+			pidFile, err := cli.NewPIDFile(pidFilePath)
+			if err != nil {
+				return err
+			}
+			err = pidFile.Acquire()
+			if err != nil {
+				return err
+			}
+			ctx := context.WithValue(cmd.Context(), config.OptPIDFile, pidFile)
```
philandstuff, line 62: I don't love using this constant for both the viper name and the context key.

also I think context keys should have user-defined types, not be strings?

philandstuff: actually I think I'd almost rather store this in a global variable at the top of this file, it feels like a lesser sin

cmd/root/root.go
```diff
@@ -50,6 +50,25 @@ func GetCommand() *cobra.Command {
...
+			cmd.SetContext(ctx)
+			return nil
+		},
+		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+			pidFile := cmd.Context().Value(config.OptPIDFile).(*cli.PIDFile)
+			err := pidFile.Release()
+			if err != nil {
+				return err
+			}
 			return nil
```
philandstuff, line 72:
```suggestion
			return := pidFile.Release()
```

cmd/root/root.go
```diff
@@ -77,6 +96,18 @@ func GetCommand() *cobra.Command {
...
 }
 
+// defaultPidFilePath returns the default path for the PID file. Notably modern OS X variants
+// do not have /var/run/
+func defaultPidFilePath() string {
+	// If we're on OS X or Windows, use the user's home directory
+	// Otherwise, use /run
+	path := "/run/pget.pid"
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
+		path = os.Getenv("HOME") + "/.pget.pid"
```
philandstuff, line 106: https://pkg.go.dev/os#UserHomeDir is slightly nicer I think, although you have to check for err. Or maybe even os.UserCacheDir?

pkg/cli/pid.go
```diff
@@ -0,0 +1,58 @@
...
+		func() error { return syscall.Flock(p.fd, syscall.LOCK_UN) },
+		p.file.Close,
+		func() error { return os.Remove(p.file.Name()) },
+	}
+	return p.executeFuncs(funcs)
+}
+
+func (p *PIDFile) writePID() error {
+	pid := os.Getpid()
+	_, err := p.file.Write([]byte(fmt.Sprintf("%d", pid)))
```
philandstuff, line 47:
```suggestion
	_, err := p.file.WriteString(fmt.Sprintf("%d", pid))
```
(double check this compiles but I think it's available here)

philandstuff (approved):
other than comments, this looks good

cmd/root/root.go
```diff
@@ -77,6 +91,18 @@ func GetCommand() *cobra.Command {
 	return cmd
 }
 
+// defaultPidFilePath returns the default path for the PID file. Notably modern OS X variants
+// do not have /var/run or /run
```
tempusfrangit, line 95:
```suggestion
// have permissioning difficulties in /var/run etc.
```

PR #127, opened 2023-12-19
Make Domains to Cache configurable

tempusfrangit:
Make domains to cache configurable via ENV. The implements the framework to also limit caching to certain prefixes to the path.

### DIFF
--- a/cmd/multifile/multifile.go
+++ b/cmd/multifile/multifile.go
@@ -138,8 +138,8 @@ func multifileExecute(ctx context.Context, manifest pget.Manifest) error {
 	// TODO DRY this
 	if srvName := config.GetCacheSRV(); srvName != "" {
 		downloadOpts.SliceSize = 500 * humanize.MiByte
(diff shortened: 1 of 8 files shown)


tempusfrangit:
What is still required is some testing and I'd like to revisit the way we talk to the CH hosts.

See: https://github.com/replicate/cluster/pull/1284 <-- This needs work to use current L2 cache, It is an example of how it could work straight to origin.

This allows us to send the URI as:
/r8/<host>/path which means we only need to update the map or in some cases the rewrite rule to enable more domains (and it could cache *any* domain not just GCS with ease).

tempusfrangit:
This PR is missing prefix matching tests. I would also like to consider not setting a default in GetDomtainsToCache. Also GetDomainsToCache is a bad function name, I'd like it to be CacheDomains to match the option.

tempusfrangit:
Texts still need to be fixed. Additionally work to warm the cache using the new scheme will be needed prior to releasing this

PR #130, opened 2023-12-19
Implement UPX compression

tempusfrangit:


tempusfrangit:
It reduces the resulting binary to 2.6MB. Still a far cry from being something we can embed in a config map. But a whole lot smaller. I wonder if we could get down to well under 1MB reliably.

.github/workflows/ci.yaml
```diff
@@ -85,6 +85,10 @@ jobs:
        name: "Checkout"
        with:
          fetch-depth: 0
+     - name: Install UPX
+       uses: crazy-max/ghaction-upx@v3
+       with:
+         install-only: true
```
tempusfrangit, line 91: I would like a better way to get UPX to the system.

PR #131, opened 2023-12-20
Improve PIDFile locking

tempusfrangit:
Instead of using flock, use exclusive create. This method has the blind spot that if the pid file has been left after a failed run, or crashed process, it is not possible to automatically clean up.

Likely it will make sense to add in an addtinal check "does pid exist from file if file exists" and "verify if pid is a pget process". If it's a stale pid we can safely remove and try the O_EXCL create...

### DIFF
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tempusfrangit:
I think this needs tests before we land it. I'll circle up on that tomorrow.

cmd/root/root.go
```diff
@@ -82,21 +83,17 @@ func defaultPidFilePath() string {
 	// Otherwise, use /run
 	path := "/run/pget.pid"
 	if runtime.GOOS == "darwin" {
-		path = os.Getenv("HOME") + "/.pget.pid"
+		path = filepath.Join(os.Getenv("HOME"), ".pget.pid")
```
philandstuff, line 86: os.UserHomeDir would be slightly nicer. Or XDG_RUNTIME_DIR if it’s set (but not sure if that’s a macOS thing)

pkg/cli/pid.go
```diff
@@ -3,52 +3,80 @@
...
-func NewPIDFile(path string) (*PIDFile, error) {
-	file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0644)
+func (p *PIDFile) tryCreateLockFile(path string) (*os.File, error) {
+	logger := logging.GetLogger()
+
+	lockedFile, err := os.OpenFile(p.Path, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
+
 	if err != nil {
-		return nil, err
+		if errors.Is(err, os.ErrExist) {
```
philandstuff, line 26: You log the error here but then you return the lock file with no error  so the caller has no idea that this failed.

philandstuff:
I’m not inclined to rush to get this landed. The current solution works and this adds more complexity.

tempusfrangit:
Totally fair. We can revisit at anytime.

Issue #128, opened 2023-12-19
Error: open /run/pget.pid: permission denied


yorickvP:
pget now requires root to run. Not very handy when testing.

$ pget version
Error: open /run/pget.pid: permission denied

$ sudo pget version
pget Version 0.5.4()

nateraw:
Just ran into this myself when updating pget in this project: https://github.com/replicate/cog-huggingface-template/pull/15. Would be very very nice to not have to run as root every time.

tempusfrangit:
There is a CLI option to change the pid path. But we should figure out a better place for the pid file if this is causing issues.

`--pid-file`; however the option is hidden from the help page. This should get you unblocked for now.

yorickvP:
Maybe default to $XDG_RUNTIME_DIR if it's set?

tempusfrangit:
@yorickvP want to contribute that fix? Happy to accept it.

tempusfrangit:
Closing this issue. @yorickvP feel free to re-open and spin up a PR with the requested change.

PR #135, opened 2023-12-28
Unhide --pid-file option

tempusfrangit:
Do not hide --pid-file option from the help page

Partial: #128


tempusfrangit:
@yorickvP @nateraw If you have additional thoughts, I am open to how we can improve this. I've unhidden the --pid-file option with this PR, but it doesn't solve the underlying "how do we make this better".

We do use `~/.pget.pid` on OS X maybe we should default to that everywhere?

PR #136, opened 2023-12-28
WIP: Implement Aliasing

tempusfrangit:
Implement aliasing, rewriting a URI Prefix to a target. This can be used to maintain the same consistent hashing jump profile for domains that map to the same cache-keys.

Example:

Target: cdn.example.com/weights
Alias:  origin.domain.org/llm-designer-name

In the above example it would be desireable to rewrite the alias to Target. Even though it is possible for the cache server should consider these equivalent due to it's internal config, it will result in different Jump hashing and therefore more content needing to be cached across more nodes in the consistent-hashing mode.

This commit does not implment additional testing to be completed in a follow-on commit.


tempusfrangit:
Really not sure I like this code. That said, it will allows us to simplify the nginx cache config if we so desire.

tempusfrangit:
I'm going to go ahead and close this PR we can revisit some other day.

PR #139, opened 2024-01-04
Fix goreleaser asset name template

zeke:
This PR updates the goreleaser config, and should solve two issues:

- Remove trailing whitespace in ilename which was causing goreleaser to prepend `default.` to some asset names. See https://github.com/goreleaser/goreleaser/issues/4513#issuecomment-1875312372
- Capitalize OS names like `Darwin` and `Linux` so you can use `uname -s` to construct OS-specific download URLs.

cc @anotherjesse


tempusfrangit (approved):

tempusfrangit:
We'll need to address the new naming where appropriate within cluster for the next version. That said, this is worth changing.

PR #148, opened 2024-01-20
WIP: VMSplice/STDOUT consumer

tempusfrangit:


tempusfrangit:
@technillogue see if this works for you, we chould generally order the requests for slices.

.goreleaser.yaml
```diff
@@ -19,7 +19,7 @@ builds:
 archives:
   - format: binary
     name_template: >-
-      {{ .ProjectName }}_{{ title .Os }}_
+      {{ .ProjectName }}_{{ tolower .Os }}_
```
tempusfrangit, line 22:
```suggestion
      {{ .ProjectName }}_{{ title .Os }}_
```

tempusfrangit: this was not an intended change.

technillogue:
this could definitely work! would it work with multifile and writing multiple files to stdout in order?

tempusfrangit:
We can make that happen! Wouldn't be too hard. This was a first pass to see if it looked good.

tempusfrangit:
It turns out the only reason we don't order downloads which causes you the problem is a historical reason for using a map instead of a list. 

Additionally there is the maximum concurrent files option already that could limit concurrency -- additionally we can set it to automatically use a max file concurrency of one for stdout/vmsplice modes.

technillogue:
how would I invoke this?

tempusfrangit:
@technillogue Rebased. Download order is now guaranteed. You will want to make sure to use `--max-concurrent-files=1` and `-o stdout` or `-o vmsplice`

technillogue:
this doesn't build

```
Error: pkg/consumer/linux_vmsplice.go:15:18: cannot use &VMSpliceConsumer{} (value of type *VMSpliceConsumer) as Consumer value in variable declaration: *VMSpliceConsumer does not implement Consumer (wrong type for method Consume)
		have Consume(io.Reader, string, int64) error
		want Consume(io.Reader, string) error
Error: pkg/consumer/stdout.go:9:18: cannot use &StdoutConsumer{} (value of type *StdoutConsumer) as Consumer value in variable declaration: *StdoutConsumer does not implement Consumer (wrong type for method Consume)
		have Consume(io.Reader, string, int64) error
		want Consume(io.Reader, string) error
```

PR #152, opened 2024-02-07
Make O_TRUNC use better

tempusfrangit:
Only use if --force is actually specified

### DIFF
--- a/pkg/consumer/consumer.go
+++ b/pkg/consumer/consumer.go
@@ -4,4 +4,5 @@ import "io"
 
 type Consumer interface {
 	Consume(reader io.Reader, destPath string) error
+	SetOverwriteTarget(force bool)
 }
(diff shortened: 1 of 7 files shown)


tempusfrangit:
This did reduce the delta (minimally) between tar and non-tar mode. I'm not certain this is worth it. @philandstuff thoughts?

pkg/consumer/consumer.go
```diff
@@ -4,4 +4,5 @@ import "io"
 
 type Consumer interface {
 	Consume(reader io.Reader, destPath string) error
+	SetOverwriteTarget(force bool)
```
philandstuff, line 7: eeh, I don't like this interface. We should pass a slice/varargs of ConsumeOptions (of which OverwriteTarget would be one instance) into Consume() instead.

philandstuff:
>  This did reduce the delta (minimally) between tar and non-tar mode. I'm not certain this is worth it. @philandstuff thoughts?

I think if we can tidy up the interface then this is a good idea. I'd also like to see us do the partial file download thing, where we download to a temp file and move to the target once the file is fully written

tempusfrangit:
Superseded by #158

PR #155, opened 2024-02-13
Gracefully handle missing SRV records

philandstuff:
This code did not communicate its intentions well.

ConsistentHashingMode has a CacheHosts member which corresponds to the SRV records for the consistent hash pods, in the correct order.  This slice can contain gaps of empty string where there are gaps in the corresponding SRV records. For example, if a statefulset `foo` has 4 pods but foo-2 is not currently ready, CacheHosts will contain this slice:

- 0: "foo-0"
- 1: "foo-1"
- 2: ""
- 3: "foo-3"

The code had a guard to ensure that it only overrides the request Host and Scheme if the CacheHost exists.  This meant that if a cache host was not ready, the request would be made directly against upstream.

This interacted poorly with CacheUsePathProxy, which also munged the HTTP path.

This commit changes the `cacheHost[i] == ""` case to treat it as a regular failure, and to perform fallbacks as necessary.  This is more correct in a few ways:

- if a single cache pod is not ready, we should retry against another cache pod before going upstream, but the current code goes straight to upstream
- it handles CacheUsePathProxy better, by ensuring we only ever munge the path if we're also sending the request to a real cache host

I've written a test for this case. It's not ideal in that in the failure case it makes a request upstream to https://weights.replicate.delivery/hello.txt. Ideally we'd spin up an origin server in the test, but all the hashes depend on the origin URL and the current way we spin up test servers chooses a new port each time.  I'd welcome improvements but it's good enough for now.

### DIFF
--- a/pkg/download/consistent_hashing_test.go
+++ b/pkg/download/consistent_hashing_test.go
@@ -310,6 +310,49 @@ func TestConsistentHashRetries(t *testing.T) {
 	assert.Equal(t, "3344761726165516", string(bytes))
 }
 
+func TestConsistentHashRetriesMissingHostname(t *testing.T) {
+	hostnames := make([]string, len(testFSes))
+	for i, fs := range testFSes {
+		ts := httptest.NewServer(http.FileServer(http.FS(fs)))
+		defer ts.Close()
+		url, err := url.Parse(ts.URL)
+		require.NoError(t, err)
+		hostnames[i] = url.Host
+	}
+	// we want to test that we never fall back to origin. So we set origin to be
+	// this canary file and ensure we don't get any data from it
+	origin := "https://weights.replicate.delivery/hello.txt"
+
+	// we deliberately "break" this cache host to make it as if its SRV record was missing
+	hostnames[0] = ""
+
+	opts := download.Options{
+		Client:               client.Options{},
+		MaxConcurrency:       8,
+		MinChunkSize:         1,
+		CacheHosts:           hostnames,
+		CacheableURIPrefixes: makeCacheableURIPrefixes("https://weights.replicate.delivery"),
+		SliceSize:            1,
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	strategy, err := download.GetConsistentHashingMode(opts)
+	require.NoError(t, err)
+
+	reader, _, err := strategy.Fetch(ctx, origin)
+	require.NoError(t, err)
+	bytes, err := io.ReadAll(reader)
+	require.NoError(t, err)
+
+	// with a functional hostnames[0], we'd see `5"37132251231713`, where the
+	// `"` character comes from upstream, but instead we should fall back to
+	// this. Note that each 0 value has been changed to a different index; we
+	// don't want every request that previously hit 0 to hit the same new host.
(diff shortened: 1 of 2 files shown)


tempusfrangit (approved):
I participated in this code by pairing with phil. This looks good and should handle the case we need to.

Additionally it removes an edge case where we were silently fallingback to origin even outside of the path-proxy code.

pkg/download/consistent_hashing_test.go
```diff
@@ -310,6 +310,49 @@ func TestConsistentHashRetries(t *testing.T) {
...
+	}
+	// we want to test that we never fall back to origin. So we set origin to be
+	// this canary file and ensure we don't get any data from it
+	origin := "https://weights.replicate.delivery/hello.txt"
+
+	// we deliberately "break" this cache host to make it as if its SRV record was missing
+	hostnames[0] = ""
+
+	opts := download.Options{
+		Client:               client.Options{},
```
nickstenning, line 330: Purely responding to your comment about this test making an upstream request: it might be nice to allow passing in a `Transport` in `client.Options`. If specified, it would be used instead of [the transport that's constructed here](https://github.com/replicate/pget/blob/451703749bb0b6ae46b6aa89fcc0cd7c1577a288/pkg/client/client.go#L61-L70).

Then you can pass in `httpmock.DefaultTransport` and use `httpmock` to intercept and/or set expectations around upstream requests.

pkg/download/consistent_hashing_test.go
```diff
@@ -310,6 +310,49 @@ func TestConsistentHashRetries(t *testing.T) {
...
+	require.NoError(t, err)
+
+	reader, _, err := strategy.Fetch(ctx, origin)
+	require.NoError(t, err)
+	bytes, err := io.ReadAll(reader)
+	require.NoError(t, err)
+
+	// with a functional hostnames[0], we'd see `5"37132251231713`, where the
+	// `"` character comes from upstream, but instead we should fall back to
+	// this. Note that each 0 value has been changed to a different index; we
```
nickstenning, line 351: It looks like this comment comes from the test above, where "each 0 value" makes sense in context because the expected value before breaking host 0 contains zeroes. I'm not quite sure what it means here?

nickstenning (approved):
Couple of comments about the tests but LGTM.

pkg/download/consistent_hashing_test.go
tempusfrangit: Absolutely. All of the tests should have an origin that shows we missed and went to origin.

pkg/download/consistent_hashing_test.go
tempusfrangit: It's related to the cache-pod index "0", where hostnames[0] is no longer valid so DoRequest will rewrite to a different index. The comment could be improved on that front.

philandstuff: It’s absolutely a copy paste fail

pkg/download/consistent_hashing_test.go
tempusfrangit: added issue #156 specifically for this comment.

philandstuff: Passing in a transport makes a lot of sense, great suggestion

PR #158, opened 2024-02-15
Implement zip extract

tempusfrangit:
Implements zip extraction consumer. This is require to handle dreambooth updates.

In golang extraction of tar and zip are within ~300ms for equivalent files (test case for dreambooth processing), lets enable lightening the load and directly handle zip files in pget.

* Include FileSize to call to consume() to support consumers like ZipExtractor that requires a size
* Implement a ReaderAt interface that can convert the muiltiChanReader, ZipExtractor requires an io.ReaderAt interface implementation
* Implement the ZipExtractor consumer
* Add some better logging when -x is supplied
* Handle overwrite support in the Consumer
* Address an Import loop.
* Add contentType as input to consumer (and output from Fetch)
* Support tar and zip extraction from multifile mode.
* Correct issue with MultiReader, it was not blocking on bufferedReader ready signal
* Implement `--unzip` (`-u`) option for setting unzip consumer
* Multifile uses `--unzip` and `--extract` options now
* README update
* Debug Logging update for unzip and tar extract


tempusfrangit:
Still Missing: muiltifile support for .tar and .zip

pkg/extract/zip.go
```diff
@@ -0,0 +1,116 @@
...
+		Str("destDir", destPath).
+		Msg("Extract")
+	zipReader, err := zip.NewReader(reader, size)
+	if err != nil {
+		return fmt.Errorf("error creating zip reader: %w", err)
+	}
+
+	for _, file := range zipReader.File {
+		err := handleFileFromZip(file, destPath, overwrite)
+		if err != nil {
```
dkhokhlov, line 35: extracted files do not end up outside the intended destination directory?
I assume ZIP code checks archive's size, structure for signs of corrupted/junk archive...
same for other arch types.

tempusfrangit: Malicious Tar and Zip checking should be added.

PR #165, opened 2024-02-19
Guarantee Order of Downloads in multifile mode

tempusfrangit:
Refactor the manifest to ensure that the order is guaranteed when downloading via multifile. Some future consumers will require guaranteed ordering. This also reduces complexity as the schemeHost is no longer needed.

### DIFF
--- a/cmd/multifile/manifest.go
+++ b/cmd/multifile/manifest.go
@@ -76,9 +76,14 @@ func parseManifest(file io.Reader) (pget.Manifest, error) {
 		if line == "" {
 			continue
 		}
-		urlString, dest, err := parseLine(line)
+		url, dest, err := parseLine(line)
 		if err != nil {
 			return nil, err
+		}
+
+		if _, err := netUrl.Parse(url); err != nil {
+			return nil, err
+
 		}
 
 		// THIS IS A BODGE - FIX ME MOVE THESE THINGS TO PGET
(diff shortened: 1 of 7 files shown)


cmd/multifile/manifest.go
```diff
@@ -61,21 +60,9 @@ func checkSeenDests(seenDests map[string]string, dest string, urlString string)
...
-	}
-
-	return manifestMap
-}
-
 func parseManifest(file io.Reader) (pget.Manifest, error) {
-	seenDests := make(map[string]string)
-	manifestMap := make(pget.Manifest)
+	seenDestinations := make(map[string]string)
+	manifest := make(pget.Manifest, 0)
```
philandstuff, line 65:
```suggestion
	manifest := make(pget.Manifest)
```
think this was added by mistake?

cmd/multifile/multifile.go
```diff
@@ -169,9 +169,7 @@ func multifileExecute(ctx context.Context, manifest pget.Manifest) error {
 	return nil
 }
 
-func numEntries(manifest pget.Manifest) (totalEntries int) {
-	for _, entries := range manifest {
-		totalEntries += len(entries)
-	}
-	return
+func numEntries(manifest pget.Manifest) int {
+	return len(manifest)
```
philandstuff, line 173: this function does nothing, get rid

cmd/multifile/manifest.go
philandstuff: ah I missed the type had changed from map to slice

cmd/multifile/manifest_test.go
```diff
@@ -49,58 +49,49 @@ func TestCheckSeenDests(t *testing.T) {
...
+	err = checkSeenDestinations(seenDests, "/tmp/file1.txt", "https://example.com/file2.txt")
 	assert.Error(t, err)
 
 	// the same destination with the same URL is also not fine
-	err = checkSeenDests(seenDests, "/tmp/file1.txt", "https://example.com/file1.txt")
+	err = checkSeenDestinations(seenDests, "/tmp/file1.txt", "https://example.com/file1.txt")
 	assert.Error(t, err)
 }
 
 func TestAddEntry(t *testing.T) {
```
philandstuff, line 64: What is this test for? What is it testing? Can it have a better name?

Also, if it's testing AddEntry, it should be in the package where AddEntry is defined

pkg/pget.go
```diff
@@ -25,14 +26,24 @@ type Options struct {
...
-type Manifest map[string][]ManifestEntry
+func (m ManifestEntry) URL() string {
+	return m.parsedURL.String()
+}
+
+// A Manifest is a slice of ManifestEntry, with a helper method to add entries
+type Manifest []ManifestEntry
+
+func (m Manifest) AddEntry(url string, destination string) (Manifest, error) {
+	parsed, err := netUrl.Parse(url)
```
philandstuff, line 41: why does AddEntry do the parsing? This feels like a bad abstraction.

philandstuff: also, the only thing we do with this URL is call String() on it so why do we need to parse it at all? Why did we change it from string?

philandstuff (commented):
overall this change looks good and simplifies things. I have a number of questions on the specifics though

cmd/multifile/manifest.go
```diff
@@ -94,24 +86,35 @@ func parseManifest(file io.Reader) (pget.Manifest, error) {
...
 			err = cli.EnsureDestinationNotExist(dest)
 			if err != nil {
 				return nil, err
 			}
 		}
-		schemeHost, err := client.GetSchemeHostKey(urlString)
-		if err != nil {
-			return nil, fmt.Errorf("error parsing url %s: %w", urlString, err)
-		}
+		if valid, err := validURL(urlString); !valid {
```
dkhokhlov, line 107: should validURL be called before seenDestinations is updated.?

tempusfrangit: Shouldn't matter order here as valid url is an error case that exits pget. If it was a non-fatal error, I think it is more reasonable to call it first. 

I am open to making it non-fatal. But that is outside the scope of this PR

PR #168, opened 2024-02-23
Enhanced MultiReader

tempusfrangit:
Build an enhanced multireader based upon the channel multireader but
support ReadAt and some other convienence bits such as ReadByte and
peek as a lead in for a better implementation for unzip, gzip, and other
fo4rmats.

This commit adds:

* BufferedMultiReader
* BufferedMultiReader Tests
* BufferedReader
* BufferedReader Tests
* Use of new BufferedMultiReader by Download package

Removal
* Old Channel MultiReader
* Old BufferedReader


nickstenning:
I've started reviewing this, and so far I have no real idea what I'm looking at. The PR description says what's in the code, but it would be really helpful for it to say _why_. Why is this needed? What problem are we solving?

pkg/download/buffer.go
```diff
@@ -65,19 +67,34 @@ type firstReqResult struct {
...
+	if errors.Is(err, io.EOF) {
+		reader.Err = fmt.Errorf("error reading response for %s: %w", resp.Request.URL.String(), err)
+		return reader.Err
+	}
+	if n != expectedBytes {
+		reader.Err = fmt.Errorf("downloaded %d bytes instead of %d for %s", n, expectedBytes, resp.Request.URL.String())
+		return reader.Err
+	}
+	return nil
+}
```
nickstenning, line 83: I think I must be missing some important context. Why do we need our own custom type/implementation if this is what we're doing with it. This appears to be reading the entire body into a `bytes.Buffer` in one go and then checking the length. At that point how is it different from:

```golang
func readBody(buf *bytes.Buffer, resp *http.Response) error {
	n, err := buf.ReadFrom(resp.Body)
 	if err != nil {
		return err
	}
	if n != resp.ContentLength {
		return fmt.Errorf("downloaded %d bytes instead of %d for %s", n, expectedBytes, resp.Request.URL.String())
	}
	return nil
}
```

(By the by, `BufferedReader.ReadFrom` effectively delegates to `bytes.Buffer.ReadFrom`, which means it cannot return `io.EOF`.)

(Oh, and this code also just completely ignores the case where `err != nil`, which doesn't seem safe.)

pkg/download/buffer.go
```diff
@@ -65,19 +67,34 @@ type firstReqResult struct {
...
 	firstReqResultCh := make(chan firstReqResult)
 	m.queue.submit(func() {
 		m.sem.Go(func() error {
 			defer close(firstReqResultCh)
-			defer br.done()
+			defer br.Done()
 			firstChunkResp, err := m.DoRequest(ctx, 0, m.minChunkSize()-1, url)
 			if err != nil {
-				br.err = err
+				br.Err = err
```
nickstenning, line 97: I'm confused by this. Are you using the `BufferedReader`'s `Err` field as an indirect signalling mechanism allowing `readBody` to return that error? If so, that seems... wrong. Why don't we just return an error here?

nickstenning (commented):
I think we should probably talk this through. I'm afraid without any context on what problem this is solving, it's really hard to evaluate whether it solves that problem.

As it is, I'm not convinced that this is doing anything that couldn't be achieved with some combination of `bytes.Buffer`, `bytes.Reader`, and possibly `bufio`.

PR #170, opened 2024-02-23
Handle Overwriting in the Consumer

### LINKED ISSUE #134: PGET -x doesn't replace symlinks/files properly errors even if -f is used.

tempusfrangit:
Make the consumer handle overwriting explicitly. This addresses edge cases with tar when extracting files.

Closes #134

### DIFF
--- a/pkg/consumer/consumer.go
+++ b/pkg/consumer/consumer.go
@@ -4,4 +4,5 @@ import "io"
 
 type Consumer interface {
 	Consume(reader io.Reader, destPath string) error
+	EnableOverwrite()
 }
--- a/pkg/consumer/consumer.go
+++ b/pkg/consumer/consumer.go
@@ -4,5 +4,4 @@ import "io"
 
 type Consumer interface {
 	Consume(reader io.Reader, destPath string) error
-	EnableOverwrite()
 }
--- a/pkg/extract/tar_test.go
+++ b/pkg/extract/tar_test.go
@@ -12,11 +12,11 @@ import (
 
 func TestCreateLinks(t *testing.T) {
 	tests := []struct {
-		name                 string
-		links                []*link
-		expectedError        bool
-		overwrite            bool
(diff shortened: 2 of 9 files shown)


tempusfrangit:
This needs to have the multi reader commit removed before merge.

pkg/consumer/consumer.go
```diff
@@ -4,4 +4,5 @@ import "io"
 
 type Consumer interface {
 	Consume(reader io.Reader, destPath string) error
+	EnableOverwrite()
```
philandstuff, line 7: This shouldn't go on the Consume interface, it should go in configuration options for the consumer when we construct it.

pkg/extract/tar_test.go
```diff
@@ -12,9 +12,11 @@ import (
...
 func TestCreateLinks(t *testing.T) {
 	tests := []struct {
-		name          string
-		links         []*link
-		expectedError bool
+		name                 string
+		links                []*link
+		expectedError        bool
+		overwrite            bool
+		createOverwritenFile bool
```
philandstuff, line 19: `Overwritten`?

pkg/extract/tar_test.go
```diff
@@ -124,3 +163,118 @@ func assertSymlinkTarget(t *testing.T, oldName, newName string) {
...
+			header: &tar.Header{
+				Name: "../invalid_file",
+			},
+			destDir:       "/tmp/valid_dir",
+			expectedError: "archive (tar) file contains file (/tmp/invalid_file) outside of target directory: ",
+		},
+		{
+			description: "directory traversal with valid file",
+			header: &tar.Header{
+				Name: "./../../tmp/valid_dir/valid_file",
```
philandstuff, line 193: I'd be happy with this test case failing tbh. I don't mind either way

philandstuff (approved):

PR #171, opened 2024-02-23
Add fileSize to the consume() call

tempusfrangit:
This will allow consumers such as unzip to utilize the data without needing to inspect the reader.


tempusfrangit:
No longer needed.

PR #173, opened 2024-03-04
fixed chunk size

philandstuff:
We have a fair amount of complexity (EqualSplit etc) involved in dynamically splitting files down into chunks based on available concurrency.

We could be less clever and use a fixed chunk size, with a fixed buffer size to go with it.  This would dovetail nicely with a sync.Pool of fixed-size buffers, and also with a configuration option to limit the total allocated RAM (implemented as a semaphore guarding how many buffers can be fetched from the sync.Pool).

This commit only does consistent hashing mode; we should do buffer mode as well if we do this work.


dkhokhlov:
side note:
https://github.com/replicate/pget/pull/173/files#diff-ec1a0e746b82d63c33a617e0b8dd205d92ac3cec2213443a6e29af27aaa37f28R139-R145
checks for one var (firstReqResult.err) but uses another var (err) in diag.

pkg/download/buffer.go
```diff
@@ -105,25 +105,22 @@ func (m *BufferMode) Fetch(ctx context.Context, url string) (io.Reader, int64, e
...
-	if fileSize <= m.minChunkSize() {
+	if fileSize <= m.chunkSize() {
 		// we only need a single chunk: just download it and finish
 		return br, fileSize, nil
 	}
 
-	remainingBytes := fileSize - m.minChunkSize()
-	numChunks := int(remainingBytes / m.minChunkSize())
+	remainingBytes := fileSize - m.chunkSize()
+	numChunks := int(remainingBytes / m.chunkSize())
```
philandstuff, line 114: this isn't right; it's the right calculation for minChunkSize (where we want to guarantee each chunk is _at least_ minChunkSize big, but it's the wrong calculation for chunkSize (where chunkSize is an _upper bound_ on the size of a chunk so we don't grow our buffers too big)

philandstuff: Took me a while to work out why our tests weren't failing but I got a failing test case in 3254d15ab8af7d32787d04c4dbfc9eaef18329b5. I'll push a fix later.

README.md
```diff
@@ -116,10 +116,10 @@ https://example.com/music.mp3 /local/path/to/music.mp3
   - Log level (debug, info, warn, error)
   - Type: `string`
   - Default: `info`
-- `-m`, `--minimum-chunk-size string`
-  - Minimum chunk size (in bytes) to use when downloading a file (e.g. 10M) 
+- `-m`, `--chunk-size string`
+  - Minimum chunk size (in bytes) to use when downloading a file (e.g. 10M)
```
philandstuff, line 120:
```suggestion
  - Chunk size (in bytes) to use when downloading a file (e.g. 10M)
```

README.md
```diff
@@ -100,10 +100,10 @@ https://example.com/music.mp3 /local/path/to/music.mp3
   - Type `Integer`
 
 ### Global Command-Line Options
-- `--max-chunks`
-    - Maximum number of chunks for downloading a given file
-    - Type: `Integer`
-    - Default: `4 * runtime.NumCPU()`
+- `--concurrency`
+  - Maximum number of chunks for downloading a given file
```
philandstuff, line 104:
```suggestion
  - Maximum number of chunks to download in parallel for a given file
```
Why is this a per-file setting? That makes no sense to me.

cmd/root/root.go
```diff
@@ -199,7 +202,7 @@ func runRootCMD(cmd *cobra.Command, args []string) error {
 // rootExecute is the main function of the program and encapsulates the general logic
 // returns any/all errors to the caller.
 func rootExecute(ctx context.Context, urlString, dest string) error {
-	minChunkSize, err := humanize.ParseBytes(viper.GetString(config.OptMinimumChunkSize))
+	chunkSize, err := humanize.ParseBytes(viper.GetString(config.OptMinimumChunkSize))
```
philandstuff, line 205:
```suggestion
	chunkSize, err := humanize.ParseBytes(viper.GetString(config.OptChunkSize))
```
We should still check MinimumChunkSize and adjust ChunkSize upwards to satisfy it if it's set.

pkg/download/options.go
```diff
@@ -16,10 +16,14 @@ type Options struct {
...
-	// used.
-	MinChunkSize int64
-	Client       client.Options
+	// Number of bytes per chunk. If set to zero, 125 MiB will be used.
+	ChunkSize int64
+
+	// Maximum size of buffer (in bytes) to allocate for storing partial
+	// download in memory.
+	MaxBuffer int64
+
```
philandstuff, line 25:
```suggestion
```
Oops, committed by mistake

cmd/root/root.go
tempusfrangit: I'm not clear why we are bothering to keep min chunk size. It seems... odd? Let's just specify the chunk size we want. Having two controls for the same thing is how we got into the mess of too many ways to control concurrency. 

If someone wants a chunk size of 1byte... it's silly but it's fine.

philandstuff: because pget is already deployed to many models which set it

if they set min chunk size to 500MiB we should respect that and not have a 125MiB chunk size.

tempusfrangit: I had wired up minchunk size and chunk size to the same value behind the scenes so they should be able to be set either flag but with --minchunk size deprecated. 

The exception is in prod where our value would continue to override (since we set with the config map).

philandstuff: Oops, yes I missed that. However:

- we should not refer to the deprecated name on this line, so my suggestion above still stands
- we wire both args up to the `chunkSize` variable but we never read that variable so it's not obvious to me that it matters (instead we fetch with viper). I don't have a good mental model for what happens? I'm not sure why we bind a variable if we're never reading it?
- [pflag has docs on how to alias flag names](https://pkg.go.dev/github.com/spf13/pflag#readme-mutating-or-quot-normalizing-quot-flag-names), maybe we should do that instead?

tempusfrangit: The aliasing is a bit weird with viper involved. Generally using an XXXVarP type instead of XXXP in cobra is the easiest way to alias as it backs to a pointer. I'm really ok either way. Your suggestions do absolutely apply in either case.

philandstuff: why do we have viper, again? is it so we can configure via environment variable? it's all a bit magic, I don't understand it, and i don't find the documentation helpful in the detail (like this variable aliasing thing, I don't think the documentation defines the behaviour here).

tempusfrangit: Viper is for the ENV binding to the flags. It's a lot of magic

Variable aliasing results in the last processed element winning i'll mkae sure the arguments cannot be both used. It should address the concerns. The ENV bits is a bit more work I'll make sure it's also clear (or at least documented) behavior

tempusfrangit (approved):
Approving at this point, however, @philandstuff I'd like you to also take another look at this before we land it.

cmd/root/root.go
```diff
@@ -109,6 +114,38 @@ func rootPersistentPreRunEFunc(cmd *cobra.Command, args []string) error {
...
+	//    Warning message will be emitted
+	// ** If both PGET_CHUNK_SIZE and PGET_MINIMUM_CHUNK_SIZE are set, use PGET_CHUNK_SIZE
+	//    Warning message will be emitted
+	// * If neither are set, use the default value
+
+	changedMin := cmd.PersistentFlags().Changed(config.OptMinimumChunkSize)
+	changedChunk := cmd.PersistentFlags().Changed(config.OptChunkSize)
+	if changedMin && changedChunk {
+		return fmt.Errorf("--minimum-chunk-size and --chunk-size cannot be used at the same time, use --chunk-size instead")
+	} else if !(changedMin && changedChunk) {
```
philandstuff, line 136:
```suggestion
	} else {
```
in this house, we accept the [principal of the excluded middle](https://en.wikipedia.org/wiki/Law_of_excluded_middle), so this condition is unnecessary.

philandstuff, line 136:
```suggestion
	} else {
```
in this house, we accept the [principal of the excluded middle](https://en.wikipedia.org/wiki/Law_of_excluded_middle), so this condition is unnecessary.

philandstuff: in fact, because the `if` block returns, we don't need an `else` at all

cmd/root/root.go
```diff
@@ -109,6 +114,38 @@ func rootPersistentPreRunEFunc(cmd *cobra.Command, args []string) error {
...
+		chunkSizeEnv := viper.GetString(config.OptChunkSize)
+		if minChunkSizeEnv != chunkSizeDefault {
+			if chunkSizeEnv == chunkSizeDefault {
+				logger.Warn().Msg("Using PGET_MINIMUM_CHUNK_SIZE is deprecated, use PGET_CHUNK_SIZE instead")
+				viper.Set(config.OptChunkSize, minChunkSizeEnv)
+			} else {
+				logger.Warn().Msg("Both PGET_MINIMUM_CHUNK_SIZE and PGET_CHUNK_SIZE are set, using PGET_CHUNK_SIZE")
+			}
+		}
+	}
```
philandstuff, line 147: Do we need to do something similar with PGET_MAX_CHUNKS and PGET_CONCURRENCY? We've removed the alias but we haven't done anything else to maintain the alias beyond binding both to the same variable (which you've established doesn't handle the environment variable case).

philandstuff: I'm going to assume "no" because we don't set these envvars in cluster and models are probably using the command line args

PR #174, opened 2024-03-04
Reusable buffers

tempusfrangit:
In an effort to limit the OOM scenarios within PGET this PR implements buffers that are fully reusable. This is done with a sync.Pool and builds upon the change to use fixed buffer sizes for any given invocation of PGET.

This PR does not limit overall memory use of PGET.

Partial: #2

### DIFF
--- a/pkg/download/buffered_reader.go
+++ b/pkg/download/buffered_reader.go
@@ -5,14 +5,15 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"sync"
 )
 
 // A bufferedReader wraps an http.Response.Body so that it can be eagerly
 // downloaded to a buffer before the actual io.Reader consumer can read it.
 // It implements io.Reader.
 type bufferedReader struct {
-	// ready channel is closed when we're ready to read
-	ready chan struct{}
+	ready bool
+	c     sync.Cond
 	buf   *bytes.Buffer
 	err   error
 }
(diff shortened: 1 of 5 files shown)


pkg/download/buffered_reader.go
```diff
@@ -82,8 +87,12 @@ func (p *readerPool) Get() *bufferedReader {
 func (p *readerPool) Put(br *bufferedReader) {
 	br.c.L.Lock()
 	defer br.c.L.Unlock()
+	if br.pool == nil {
+		return
+	}
 	br.ready = false
 	br.err = nil
+	br.pool = nil
 	br.buf.Reset()
```
tempusfrangit, line 96: Thinking this through, these likely should be moved up to .Close() and not be part of .Put().

pkg/download/buffered_reader.go
```diff
@@ -71,6 +71,11 @@ func (b *bufferedReader) waitOnReady() {
 	b.c.L.Unlock()
 }
 
+func (b *bufferedReader) Close() error {
+	b.pool.Put(b)
+	return nil
```
tempusfrangit, line 76:
```suggestion
	b.c.L.Lock()
	defer b.c.L.Unlock()
	b.ready = false
	b.err = nil
	b.pool = nil
	b.buf.Reset()

	b.pool.Put(b)

	return nil
```

pkg/download/buffered_reader.go
tempusfrangit, line 96:
```suggestion
	if br.pool == nil {
		return
	}
```

philandstuff:
I've looked at this and implemented an alternative in #177.

tempusfrangit:
> I've looked at this and implemented an alternative in #177.

WFM lets use #177

PR #177, opened 2024-03-06
Reuse buffers used by bufferedReader

philandstuff:
This is an alternative to #174.  The key differences are:

- we use sync.Pool to store *bytes.Buffer instances rather than bufferedReader instances
- we keep the ready channel instead of using sync.Cond
- we return the buffer to the pool in Read() rather than requiring an explicit Close()

Overall this changes fewer parts of the codebase and is (IMHO) simpler and easier to understand.

### DIFF
--- a/pkg/download/buffered_reader.go
+++ b/pkg/download/buffered_reader.go
@@ -21,6 +21,8 @@ type bufferedReader struct {
 
 var _ io.Reader = &bufferedReader{}
 
+var emptyBuffer = bytes.NewBuffer(nil)
+
 func newBufferedReader(pool *bufferPool) *bufferedReader {
 	return &bufferedReader{
 		ready: make(chan struct{}),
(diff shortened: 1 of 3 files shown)


nickstenning (approved):

pkg/download/buffered_reader.go
```diff
@@ -15,25 +16,37 @@ type bufferedReader struct {
...
 	}
-	return b.buf.Read(buf)
+	n, err := b.buf.Read(buf)
+	// If we've read all the data,
+	if b.buf.Len() == 0 {
+		// replace our buffer with something that will always return EOF on
+		// future reads
+		b.buf = bytes.NewBuffer(nil)
+		// and return the buffer to the pool
+		b.pool.Put(b.buf)
```
nickstenning, line 47:
```suggestion
                 // return the buffer to the pool
		b.pool.Put(b.buf)
		// and replace our buffer with something that will always return EOF on
		// future reads
		b.buf = bytes.NewBuffer(nil)
```

tempusfrangit (approved):
LGTM. 

I have a slight preference for the sync.Cond but that's 100% a personal preference and amounts to painting pget green. This is a great change let's use this one.

PR #178, opened 2024-03-06
Test buffered reader / Ensure EmptyBuffer does not get content in it

tempusfrangit:
This PR introduces tests for bufferedReader to ensure there are no regressions, additionally, it adds a panic() case where we've attempted to use the downloadBody or a .readFrom() on the bufferedReader when emptyBuffer is the underlying bytes.Buffer. 

The panic() is indicating a programming error, additionally it helps ensure the assumption that emptyBuffer will EOF is guaranteed. The only use of the br.buf.ReadFrom is from tests and/or downloadBody.

### DIFF
--- a/pkg/download/buffered_reader.go
+++ b/pkg/download/buffered_reader.go
@@ -22,6 +22,7 @@ type bufferedReader struct {
 var _ io.Reader = &bufferedReader{}
 
 var emptyBuffer = bytes.NewBuffer(nil)
+var errContentLengthMismatch = fmt.Errorf("Content-Length doesn't match expected bytes")
 
 func newBufferedReader(pool *bufferPool) *bufferedReader {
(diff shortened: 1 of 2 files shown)


philandstuff:
Ooof, bufferedReader sure is hard to test, isn't it? This PR made me feel bad enough about its interface that I have raised #180 to change it to have a better interface (purely on io abstractions with nothing http-specific) and also a better implementation (delegating more logic to bufio.Reader).  If we do that, we don't have to test for some of the behaviour we're testing for here - there's no longer an emptyBuffer that we might write to, we don't have to implement checks for full buffers, bufio.Reader has a fixed-size buffer so we can't unintentionally grow it, bufio.Reader handles passing errors during fetch out on the Read() interface appropriately, etc.

PR #180, opened 2024-03-07
reimplement bufferedReader on top of bufio.Reader

philandstuff:
This reimplements bufferedReader on top of bufio.Reader.

This doens't make the code shorter, but it delegates more complexity to standard library functions.  In particular, bufio.Reader:

- handles storing errors and passing them out via Read() where appropriate
- handles errors where a read would exceed the buffer size

Furthermore this commit removes any HTTP concerns from bufferedReader.  The downloadBody() function has been changed to a generic ReadFrom().

Stacked on #179.

### DIFF
--- a/pkg/download/buffered_reader.go
+++ b/pkg/download/buffered_reader.go
@@ -1,26 +1,30 @@
 package download
 
 import (
-	"bytes"
-	"fmt"
+	"bufio"
 	"io"
-	"net/http"
+	"strings"
 	"sync"
 )
 
-// A bufferedReader wraps an http.Response.Body so that it can be eagerly
-// downloaded to a buffer before the actual io.Reader consumer can read it.
-// It implements io.Reader.
+// A bufferedReader wraps a bufio.Reader so that it can be shared between
+// goroutines, with one fetching data from an upstream reader and another
+// reading the data.  It implements io.ReaderFrom and io.Reader.  Read() will
+// block until Done() is called.
+//
+// The intended use is: one goroutine calls Read(), which blocks until data is
+// ready.  Another calls ReadFrom() and then Done().  The call to Done()
+// unblocks the Read() call and allows it to read the data that was fetched by
+// ReadFrom().
 type bufferedReader struct {
 	// ready channel is closed when we're ready to read
 	ready chan struct{}
-	buf   *bytes.Buffer
+	buf   *bufio.Reader
 	pool  *bufferPool
 }
 
 var _ io.Reader = &bufferedReader{}
-
-var emptyBuffer = bytes.NewBuffer(nil)
+var _ io.ReaderFrom = &bufferedReader{}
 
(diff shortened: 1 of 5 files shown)


pkg/download/buffered_reader.go
```diff
@@ -35,57 +39,58 @@ func newBufferedReader(pool *bufferPool) *bufferedReader {
...
 
-func (b *bufferedReader) done() {
-	close(b.ready)
+func (b *bufferedReader) ReadFrom(r io.Reader) (int64, error) {
+	b.buf.Reset(r)
+	bytes, err := b.buf.Peek(b.buf.Size())
+	if err == io.EOF {
+		// ReadFrom does not return io.EOF
+		err = nil
+	}
```
tempusfrangit, line 63: While this is generally unlikely, it is possible for the peek to not completely fill the buffer before 100 iterations. This could result in a contentLength mismatch as it would return fewer than requested bytes. This likely needs to be a for {} loop and handle the case of `io.ErrNoProgress`, another error (exit) or until the buffer's filled to expected content length

https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/bufio/bufio.go;l=109

tempusfrangit (changes requested):
Generally I like this PR, however there is a concern with the read/fill in the case of a slow / hung download.

The current bytes.Buffer implementation does not suffer from a "max number of empty reads" scenario.

pkg/download/buffered_reader.go
philandstuff: if peek doesn't completely fill the buffer, we get another chance to fill the buffer when `Read` is called, because bufio.Reader is ultimately a wrapper around an underlying Reader. We are doing best effort to have the buffer full before the consumer calls Read, but if we aren't successful the behavior is still correct.

The downside is we effectively give up trying to download until the consumer gets around to reading this particular Reader. I think that's ok - this is "generally unlikely" as you say and I'm happy for it to not be 100% optimized.

tempusfrangit: Yes, additionally this code in consistent_hashing.go:

```go
					n, err := br.ReadFrom(resp.Body)
					if err != nil {
						return err
					} else if n != contentLength {
						return ErrContentLengthMismatch{downloadedBytes: n, contentLength: contentLength}
					}
```

Will not let us get to the Read during the consumer, and we still need to use the .Done() call for the channel before we read. As implemented if we hit the unlikely scenario we might deadlock the download?

philandstuff: the call do br.Done() is in a defer at the top of that block so I'm unclear what problem you're describing here?

philandstuff: I pushed a commit to do a loop checking for progress.

tempusfrangit (approved):

Issue #1, opened 2023-05-10
Enhancement Request: GZIP support (tar mode)


Pwntus:
Writing down how I implemented ZIP at first. As discussed, weights are not good for ZIP compression due to their "noisy" structure - decompressing with ZIP adds ~50-60 seconds.

1. Import `compress/gzip`.
2. When extracting, first create a uncompressed stream before feeding to the `tar` reader:

```go
func extractTarFile(buffer *bytes.Buffer, destDir string) error {
  uncompressedStream, err := gzip.NewReader(buffer)
  if err != nil {
    return err
  }

  tarReader := tar.NewReader(uncompressedStream)
  // ...
}
```

tempusfrangit:
This should be much easier to implement now, do we want it?

PR #183, opened 2024-03-12
Implement Stream Decompression for tar

tempusfrangit:
Implement stream decompression for tar files. This mean that -x becomes smart enough to handle lzw, bzip2, gzip, and xz (common tar compression) formats automatically. This will remove a sharp edge on pget and handle cases of compressed tar files elegantly. 

It is recommended to use uncompressed TAR as a weights source as decompression is significantly slower for very minimal space savings. 

This code is intended to make pget slightly more forgiving but does not improve performance.

Closes: #1


pkg/extract/compression.go
```diff
@@ -0,0 +1,137 @@
...
+		compressionByte := input[2]
+		// litWidth is guaranteed to be at least 9 per specification, the high order 3 bits of byte[2] are the litWidth
+		// the low order 5 bits are only used by non-unix implementations, we are going to ignore them.
+		litWidth := int(compressionByte>>5) + 9
+		log.Debug().
+			Str("type", "lzw").
+			Int("litWidth", litWidth).
+			Msg("Compression Format")
+		return lzwDecompressor{
+			order:    lzw.MSB,
```
dkhokhlov, line 79: lzw.MSB - is it true for all streams? i think calculation of litWidth affected by it.

tempusfrangit:
We should probably do something where we log a warning about compressed streams being a poor experience.

pkg/extract/compression.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/pierrec/lz4"
+	"github.com/ulikunitz/xz"
+
+	"github.com/replicate/pget/pkg/logging"
+)
+
+const (
+	peekSize = 8
+
+	gzipMagic = 0x1F8B
```
philandstuff, line 19: did you consider making these byte slices rather than integers? Then you don't have to do mucking around with binary.BigEndian or padding, you can just straight up use bytes.HasPrefix. eg:
```suggestion
	gzipMagic = []byte{0x1F,0x8B}
```

pkg/extract/tar.go
```diff
@@ -22,10 +22,23 @@ type link struct {
 	newName  string
 }
 
-func TarFile(reader io.Reader, destDir string, overwrite bool) error {
+func TarFile(r io.Reader, destDir string, overwrite bool) error {
 	var links []*link
 
 	startTime := time.Now()
+	peekReader := &peekReader{reader: r}
```
philandstuff, line 29:
```suggestion
	peekReader := bufio.NewReaderSize(r,8)
```
then peekReader.Peek(8) gets you the magic bytes

philandstuff (approved):
I think this could be simplified quite a bit but it's good

pkg/extract/compression.go
tempusfrangit: I remember why! Because I wanted to use constants for these vs vars -- which is silly looking back on it now.

Issue #188, opened 2024-03-15
CRITICAL BUG: Panic on waitgroup use


tempusfrangit:
There is a panic on re-use of a wait group. This must be debugged before we roll out a 0.7.x version.

tempusfrangit:
This is related to multifile use-case, where it is possible to call sem.Wait() from multiple go routines but more importantly because DownloadFile is called in goroutines from downloadFilesFromManifest, this results in the possiblity that the shared errorgroup on the downloader results in adding to the errgroup after .wait() is called. This results in a panic.

This means that a shared errgroup is not possible on the download strategy.

PR #189, opened 2024-03-15
Revert "implement Wait() on Strategy to fetch remaining errors (#179)"

tempusfrangit:
This reverts commit c72c38485b2389513b8051912b47e7f63db54641.

We cannot add more to the underlying waitgroup in eg after we call .Wait(). We're calling .Wait() as part of DownloadFile which could in some cases result in after .Wait() before return additional additions to the waitgroup. This is surfaced in multifile mode when the multifile number of buffers is below a critical threshold allowing for the downloader to begin a second file before the first file completed.

Closes: #188

### DIFF
--- a/pkg/download/buffer.go
+++ b/pkg/download/buffer.go
@@ -73,6 +73,7 @@ func (m *BufferMode) Fetch(ctx context.Context, url string) (io.Reader, int64, e
 			defer br.Done()
 			firstChunkResp, err := m.DoRequest(ctx, 0, m.chunkSize()-1, url)
 			if err != nil {
+				br.err = err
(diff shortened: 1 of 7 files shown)


tempusfrangit:
Alternately we can go to a proper semaphore or a channel-based semaphore instead of the errgroup limiter.

philandstuff:
I implemented a roll-forward fix in #190, which I'd prefer if you agree

tempusfrangit:
> I implemented a roll-forward fix in #190, which I'd prefer if you agree

Yeah that's fine. This revert was me trying to understand where the double wait is coming from. 

Let me review and we can go with the other fix instead.

PR #190, opened 2024-03-16
Only wait for errGroup once

### LINKED ISSUE #188: CRITICAL BUG: Panic on waitgroup use

philandstuff:
DownloadFile() was used as a public interface but also internally.  This meant that it was called multiple times from DownloadFiles(), resulting in calling Wait on the errgroup multiple times, which caused a panic.

This extracts a common private downloadFile() function which can be called from both public functions, and ensures we only call Downloader.Wait() once within each public function.

Fixes #188.

### DIFF
--- a/pkg/pget.go
+++ b/pkg/pget.go
@@ -37,6 +37,15 @@ func (m Manifest) AddEntry(url string, destination string) Manifest {
 }
 
 func (g *Getter) DownloadFile(ctx context.Context, url string, dest string) (int64, time.Duration, error) {
+	fileSize, totalElapsed, err := g.downloadFile(ctx, url, dest)
+	if err != nil {
+		return fileSize, totalElapsed, err
+	}
+	err = g.Downloader.Wait()
+	return fileSize, totalElapsed, err
+}
+
+func (g *Getter) downloadFile(ctx context.Context, url string, dest string) (int64, time.Duration, error) {
 	if g.Consumer == nil {
 		g.Consumer = &consumer.FileWriter{}
 	}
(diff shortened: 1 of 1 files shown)


philandstuff:
Side note: it occurs to me that we have two errGroups here, but we only set a context on one of them. We really ought to set a context on both, so that a single chunk failure causes us to abort the rest.

pkg/pget.go
```diff
@@ -99,6 +104,10 @@ func (g *Getter) DownloadFiles(ctx context.Context, manifest Manifest) (int64, t
 	if err != nil {
 		return 0, 0, fmt.Errorf("error downloading files: %w", err)
 	}
+	err = g.Downloader.Wait()
+	if err != nil {
+		return 0, 0, fmt.Errorf("error downloading files: %w", err)
+	}
```
tempusfrangit, line 110: I don't think this 100% solves the issue/panic. 

Double calls to .Wait() isn't the underlying issue. Since we execute the Fetches through g.Go() those happen in a wait group and if we call .Wait() here and deep in fetch we are slow and end up adding another element to the errgroup we'll see the same panic as before. 

Now this code does reduce the potential if the max concurrent files is < the length of the manifest as we would block on errgroup.Wait a few lines up. 

I think this points to a problem with complexity of using the errgroup as the semaphore and error channel.

tempusfrangit (commented):
See my comment I think we are still at risk of panic()

pkg/pget.go
philandstuff: you're right that there's still a risk. I'll have another go. I think we might be missing an abstraction here.

philandstuff: I've had a long think about this and I now think two things:

1. this PR actually is safe. We only call Wait() after every returned stream has been fed to consumer.Consume(), which means there really can't be any remaining goroutines reading to call errgroup.Go().
2. this is really really hard to think about, and I'm sure there's a simpler design inside screaming to get out.

So I think we should merge this PR as a way to fix #188, but I'm thinking hard about making this simpler overall.

PR #191, opened 2024-03-17
Rearrange concurrency control and error handling

### LINKED ISSUE #188: CRITICAL BUG: Panic on waitgroup use

philandstuff:
Fixes #188.

This is my second attempt after @tempusfrangit's feedback on #190.

We have too many concurrency primitives, and we have too many ways to handle errors.

In particular, in both buffer.go and consistent_hashing.go:

- there is an ErrGroup and a workQueue for submitting asynchronous work to, and they interact in nontrivial ways
- the `sem` ErrGroup is responsible for imposing concurrency control but it is also responsible for collecting errors from asynchronous goroutines.

This commit changes this all:

- the `sem` errgroup is removed
- workQueue is now responsible for concurrency control; it has a limited size worker pool.
- nearly all errors that would have been emitted to the ErrGroup are now funneled through bufferedReader instead

bufferedReader has has some significant changes:

- ReadFrom() is now Prefetch() and it no longer returns an error
- errors can be injected from outside via recordError() - this is helpful if an error that didn't come from the underlying io.Reader is still relevant to the consumer (eg if we failed to initiate an http request for a chunk, so we never created the response body reader in the first place)

Another big change is that bufferedReader no longer acquires a bufio.Reader from the bufferPool when created; instead it delays the acquisition until Prefetch() is called.

This means that it's now safe to create all the bufferedReaders a given Fetch() call will need ahead of time. Since creating a bufferedReader doesn't allocate a memory buffer, it won't cause OOM problems.  And this in turn means we can ditch chanMultiReader and use the standard library io.MultiReader.

workQueue is now a priority work queue.  The priority is because we want to prioritise finishing existing files over starting new ones; so the first chunk download happens via submitLow() and all subsequent chunk downloads happen via submitHigh().

I think long term the right place for concurrency control is the bufferPool - if the resource we actually care about is RAM, then bufferPool should guard access to memory buffers with a semaphore.  But we need to be careful of deadlocks if we do that.


tempusfrangit:
First pass this looks good but I want to spend some time reviewing it. 

This definitely approaches the problem in a much better way than we have currently.

pkg/download/buffered_reader.go
```diff
@@ -39,6 +46,16 @@ func newBufferedReader(pool *bufferPool) *bufferedReader {
 // pool.
 func (b *bufferedReader) Read(buf []byte) (int, error) {
 	<-b.ready
+	err := b.readErr()
+	if err != nil {
+		return 0, err
+	}
+	if b.buf == uninitializedReader {
```
nickstenning, line 53: More of a nit than meaningful feedback, but using a sentinel `bufio.Reader` for this case seems a bit clever. It also allocates a default-size buffer that is never used.

Perhaps you could define the `buf` field in terms of the interface you require rather than as a concrete `*bufio.Reader`. That would allow you to treat `nil` as uninitialized, and when you're done with it you can replace it with your own `alwaysReturnEOFReader{}`?

pkg/download/work_queue.go
```diff
@@ -1,26 +1,52 @@
 package download
 
-// workQueue takes work items and executes them serially, in strict FIFO order
+// workQueue takes work items and executes them serially, with n concurrent
+// workers.  It allows for a simple high/low priority split between work.  We
```
nickstenning, line 4: Maybe I'm being slow, but what does "with n concurrent workers" mean here?

pkg/download/buffer.go
```diff
@@ -64,42 +57,37 @@ type firstReqResult struct {
...
-			n, err := br.ReadFrom(firstChunkResp.Body)
-			if err != nil {
-				return err
-			} else if n != contentLength {
-				return ErrContentLengthMismatch{downloadedBytes: n, contentLength: contentLength}
-			}
-			return nil
-		})
+	m.queue.submitLow(func() {
+		defer close(firstReqResultCh)
```
nickstenning, line 64: I think I probably need to sit down with you to really grok what this code is doing, but there seems to be a pretty significant change between the old code and the new code, which is that the enqueued "work" items here actually executed in parallel before (because `m.sem.Go` returns immediately), whereas now they execute serially. Doesn't that matter?

pkg/download/buffered_reader.go
philandstuff: hmm, I think I can just use nil anyway since *bufio.Reader is a pointer type

pkg/download/work_queue.go
philandstuff: it means I forgot to put a for loop in the code 🤦 I will also rewrite the comment to make more sense

pkg/download/buffer.go
philandstuff: let's discuss.

philandstuff:
I pushed another commit. I'm so sorry, I keep doing this.

But I think I have better abstractions now.

The work queue is now responsible for managing buffers, and will only execute
work if there is a buffer available for it.  This means we can limit the size of
the buffer pool but we don't run the risk of deadlock.

bufferedReader has been rewritten to be readerPromise.  It's the promise of an
io.Reader in future.

Instead of using bufio.Reader.Peek() to prefetch data, we prefetch using
io.ReadFull().  This is a really good fit for our use case - we already have an
expected size of the data given by resp.ContentLength; if the data we get is too
short, io.ReadFull returns io.ErrUnexpectedEOF, so we don't need our own
ErrContentLengthMismatch type any more.

----

This pre-allocates all the buffers, but there's a plausible version of this code that only allocates buffers on-demand as needed (using a sync.Pool as before).

pkg/download/reader_promise.go
```diff
@@ -0,0 +1,62 @@
...
+	<-b.ready
+	if b.err != nil {
+		return 0, b.err
+	}
+	n, err := b.reader.Read(buf)
+	// If we've read all the data,
+	if err == io.EOF && b.buf != nil {
+		// unblock the producer
+		close(b.finished)
+		b.buf = nil
```
philandstuff, line 48:
```suggestion
		b.buf = nil
		b.err = io.EOF
```

tempusfrangit (approved):
Did a pairing session and walked through the code. This is mergable and we should performance test before cutting a release.

PR #234, opened 2024-12-02
Error when too few bytes are read

tempusfrangit:
In the case where we have a scenario that results in fewer than expected bytes being read from the tar compression or fewer than expected bytes written from the file writer consumer, propagate an error to the main program and exit with non-success status.

Observed behavior shows that within the tar-extraction code path in certain circumstances a clean EOF is received by the tarReader and the archive is partially extracted. This generally presents as files missing in the destination directory.

New code paths expect the correct number of bytes to be read or it will propagate a non-zero exit.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,4 +1,5 @@
 project_name: pget
+version: 2
 before:
   hooks:
     - go mod tidy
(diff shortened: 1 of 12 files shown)


meatballhat (approved):
2️⃣ ➕ 2️⃣ 🟰 5️⃣

PR #236, opened 2024-12-03
Fix PGET Expected Byte Check

tempusfrangit:
Tar files have padding. Verify the end of the file is null and only error if there is non-null or we haven't matched content-length from origin.


philandstuff (approved):
I think this could be simpler with io.ReadAll() - we wouldn't need the outer for loop - but this is fine

tempusfrangit:
> I think this could be simpler with io.ReadAll() - we wouldn't need the outer for loop - but this is fine

Good point. I'll verify that works and get an update.

tempusfrangit:
The only risk here is if someone creates a malicious tar with massive padding it could force an oomkill.

I am generally ok with that trade off.

PR #251, opened 2025-03-20
Send metrics if an endpoint exists

nevillelyh:


pkg/pget.go
```diff
@@ -123,3 +140,32 @@ func (g *Getter) downloadAndMeasure(ctx context.Context, url, dest string, total
...
+		Source: "pget",
+		Type:   "download",
+		Data:   map[string]any{"url": url, "size": size},
+	}
+	body, err := json.Marshal(payload)
+	if err != nil {
+		logger.Debug().Err(err).Any("payload", payload).Msg("Error marshalling metrics")
+		return
+	}
+	fmt.Println(string(body))
```
tempusfrangit, line 161:
```suggestion
```

pkg/pget.go
```diff
@@ -123,3 +140,32 @@ func (g *Getter) downloadAndMeasure(ctx context.Context, url, dest string, total
 	totalSize.Add(fileSize)
 	return nil
 }
+
+func (g *Getter) sendMetrics(url string, size int64) {
+	logger := logging.GetLogger()
+	endpoint, ok := os.LookupEnv(MetricsEndpointKey)
```
tempusfrangit, line 146: There is a way to do this already via VIPER. add this as part of the config options and reference the config when kicking off the `g.sendMetrics()` call

see the plumbing here: https://github.com/replicate/pget/blob/main/pkg/config/optnames.go and how we pass config down. os.LookupEnv is easier but it should be done correctly. I don't expect we'll rip this out in the future as it's OPT in.

tempusfrangit (changes requested):

pkg/pget.go
nevillelyh: Fixed. Also added `version` to `data` because why not.

pkg/pget.go
```diff
@@ -123,3 +142,31 @@ func (g *Getter) downloadAndMeasure(ctx context.Context, url, dest string, total
 	totalSize.Add(fileSize)
 	return nil
 }
+
+func (g *Getter) sendMetrics(url string, size int64) {
+	logger := logging.GetLogger()
+	endpoint := viper.GetString(config.OptMetricsEndpoint)
```
tempusfrangit, line 148: We might want to rework this to be part of the gettr options. It's fine here but also there is no need to query the ENV every time. it wont change.

pkg/pget.go
```diff
@@ -1,11 +1,19 @@
...
 	"fmt"
+	"net/http"
 	"sync/atomic"
 	"time"
 
+	"github.com/spf13/viper"
+
+	"github.com/replicate/pget/pkg/config"
+	"github.com/replicate/pget/pkg/version"
+
```
tempusfrangit, line 16: Grouping is incorrect.

viper should be in the same group as go-humanize, pget shoulc be in the grouping with consumer download etc.

tempusfrangit (approved):

PR #253, opened 2025-03-20
Imeplement Cache Support in BufferMode

tempusfrangit:
Support specifying a cache endpoint for pget to utilize. This is
distinct from the CH mode as we do not need SRV records, a static target
is sufficient.

Using the standard PROXY env variables will result in non-pget traffic
being redirected. The Cache should be explicitly opt-in.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.23.3
+golang 1.24.1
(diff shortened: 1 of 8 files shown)


tempusfrangit:
This needs to be reworked, the whole SRV record lookup isn't needed  🤦🏻

NikhilSinha1 (approved):

PR #255, opened 2025-03-28
Handle full file response

NikhilSinha1:
### Summary

Recently had some issues with `pget` failing for smaller files because they did not respond with `content-range`. Checking [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Range), it looks like there isn't an expectation that `content-range` is provided when the full file is returned along with a `200 OK`. `pget` should handle `200 OK`s without `content-range`.

### Test Plan

Unit tests added to test behavior.

### DIFF
--- a/pkg/download/consistent_hashing.go
+++ b/pkg/download/consistent_hashing.go
@@ -68,6 +68,28 @@ func (m *ConsistentHashingMode) chunkSize() int64 {
 	return chunkSize
 }
 
+func (m *ConsistentHashingMode) getFileSizeFromResponse(resp *http.Response) (int64, error) {
+	// If the response is a 200 OK, we need to parse the file size assuming the whole
+	// file was returned. If it isn't, we will assume this was a 206 Partial Content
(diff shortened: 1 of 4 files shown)


pkg/download/consistent_hashing.go
```diff
@@ -68,6 +68,28 @@ func (m *ConsistentHashingMode) chunkSize() int64 {
 	return chunkSize
 }
 
+func (m *ConsistentHashingMode) getFileSizeFromResponse(resp *http.Response) (int64, error) {
+	// If the response is a 200 OK, we need to parse the file size assuming the whole
+	// file was returned. If it isn't, we will assume this was a 206 Partial Content
+	// and parse the file size from the content range header. We wouldn't be in this
+	// function if the response was not between 200 and 300, so this feels like a
+	// reasonable assumption
```
tempusfrangit, line 76: Reasonable assertion.

tempusfrangit (approved):

PR #257, opened 2025-04-11
Fix off by 1 error

NikhilSinha1:
### Summary

Due to a misunderstanding in regex, I thought the value here would be a 0 indexed value. However, it is a 1 indexed value, so there was no need to do the subtraction by 1 that I put in here. This caused an off by 1 error when trying to write the file.

### Test Plan

Added a unit test for this. If you try to run this unit test with the old version of this code, you should see the error
```
--- FAIL: TestDownloadSmallFileWith200 (0.01s)
    /pget/pkg/pget_test.go:149: 
        	Error Trace:	/pget/pkg/pget_test.go:149
        	Error:      	Received unexpected error:
        	            	error writing file: expected 27 bytes, wrote 28
        	Test:       	TestDownloadSmallFileWith200
```

### DIFF
--- a/pkg/download/buffer.go
+++ b/pkg/download/buffer.go
@@ -56,6 +56,7 @@ func (m *BufferMode) getFileSizeFromContentLength(contentLength string) (int64, 
 	if err != nil {
 		return 0, err
 	}
+
 	return size, nil
 }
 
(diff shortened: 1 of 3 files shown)


tempusfrangit (approved):
This feels like something we can also show the behavior in a test case. While you're welcome to merge this as-is I'd like to have some guard that a future refactor doesn't break things.


✅ but I'd love to also see a test for this case.

PR #265, opened 2025-05-13
Support custom headers

nevillelyh:


NikhilSinha1 (approved):
Related to auth header work we are doing to get FUSE and hermes wired up to each other

PR #266, opened 2025-05-31
Add throughput to metrics

NikhilSinha1:
### Summary

Add throughput to metrics sent to Honeycomb

### DIFF
--- a/pkg/pget.go
+++ b/pkg/pget.go
@@ -64,18 +64,22 @@ func (g *Getter) DownloadFile(ctx context.Context, url string, dest string) (int
 	// downloadElapsed := time.Since(downloadStartTime)
 	// writeStartTime := time.Now()
 
-	// Fire and forget metrics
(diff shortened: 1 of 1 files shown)


pkg/pget.go
```diff
@@ -144,17 +148,24 @@ func (g *Getter) downloadAndMeasure(ctx context.Context, url, dest string, total
...
 	endpoint := viper.GetString(config.OptMetricsEndpoint)
 	if endpoint == "" {
 		return
 	}
 
+	data := map[string]any{"url": url, "size": size, "version": version.GetVersion()}
+	if downloadFailed {
+		data["error"] = true
+	} else {
+		data["throughput"] = throughput
```
nevillelyh, line 162: Can we use something like `bytes_per_second` to be clearer and consistent with existing graph & https://github.com/replicate/hf_transfer/pull/1?

